PR #1, opened 2017-01-19
[RFR] Bootstrap frontend

djhi:
screen
Awesome isn't it ?


ThieryMichel:
I think we should not commit built file.

PR #3, opened 2017-01-19
[RFR] Add server

ThieryMichel:


.babelrc
```diff
@@ -0,0 +1,8 @@
+{
+    "presets": [
```
djhi, line 2: You should use babel-preset-env with:
```js
"presets": [
    ["env", {
        "targets": {
            "node": "current"
        },
        "include": ["transform-es2015-spread"],
        "useBuiltIns": true
    }]
],
"plugins": ["transform-object-rest-spread"]
```
Why the stage-2 ?

src/api/launcher.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+require('babel-register');
```
djhi: Avoid requiring babel here. Use `--harmony-async-await --require babel-register` in the pm2 config

djhi (approved):

PR #15, opened 2017-01-25
[RFR] Column transform

ThieryMichel:


src/api/controller/api/publication.js
```diff
@@ -1,9 +1,9 @@
 export default async (ctx) => {
     const publishedDatasetCount = await ctx.publishedDataset.count();
-    const publishedModel = await ctx.publishedModel.find({}).toArray();
+    const fields = await ctx.field.find({}).toArray();
 
     ctx.body = {
         published: publishedDatasetCount > 0,
-        model: publishedModel,
+        model: fields,
```
djhi: we should name this accordingly in the UI too

djhi (changes requested):
I find the indentation make the code quite hard to read but appart from that, nice work !

djhi (approved):

PR #20, opened 2017-01-26
[RFR] Refactor loading

djhi:
Extracted Loading in a dedicated component

### DIFF
--- a/src/app/js/lib/Card.js
+++ b/src/app/js/lib/Card.js
@@ -8,7 +8,6 @@ const getStyles = style => ({
 
 const CardComponent = ({ children, style, ...props }) => (
     <Card
-        className="loading"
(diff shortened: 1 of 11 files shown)


src/app/js/lib/Card.js
```diff
@@ -0,0 +1,25 @@
...
+import { Card } from 'material-ui/Card';
+
+const getStyles = style => ({
+    marginTop: '0.5rem',
+    ...style,
+});
+
+const CardComponent = ({ children, style, ...props }) => (
+    <Card
+        className="loading"
```
ThieryMichel: should'nt it take className from props?

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #21, opened 2017-01-26
[RFR]Field api

ThieryMichel:
follow #15

### DIFF
--- a/src/api/models/field.spec.js
+++ b/src/api/models/field.spec.js
@@ -1,0 +1,82 @@
+import expect from 'expect';
+import { validateField } from './field';
+
+describe.only('field', () => {
+    describe('validateField', () => {
+        it('should return field if valid', () => {
+            const field = {
+                name: 'name',
+                label: 'label',
+                transformers: [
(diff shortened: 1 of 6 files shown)


src/api/models/field.spec.js
```diff
@@ -0,0 +1,82 @@
+import expect from 'expect';
+import { validateField } from './field';
+
+describe.only('field', () => {
```
djhi: only spotted

djhi (changes requested):

src/api/services/mongoClient.js
```diff
@@ -9,6 +9,7 @@ export default async (ctx, next) => {
     ctx.dataset = dataset(ctx.db);
     ctx.publishedDataset = publishedDataset(ctx.db);
     ctx.field = field(ctx.db);
+    await ctx.field.createIndex({ name: 1 }, { unique: true });
```
djhi: This should go in the field model IMO

djhi (changes requested):

djhi (approved):

PR #26, opened 2017-01-30
[RFR] Handle EzMaster configuration

djhi:
- [x] Read ezMaster config and put it in koa context
- [x] Updated login to check credentials against those inside the ezMaster config


package.json
```diff
@@ -28,6 +28,8 @@
     "classnames": "2.2.5",
     "config": "1.24.0",
     "csv-parse": "1.1.10",
+    "expect": "1.20.2",
+    "inist-ark": "2.0.0",
```
djhi, line 32: I was planning to use it to validate the config but it wasn't possible. I kept it anyway as we will use it later to generate ARK uris.

PR #31, opened 2017-01-31
[RFR] Link function

djhi:
- [x] Refactored transformations to introduce context
- [x] Applied transformations by stages (pre, post)
- [x] Implemented `LINK` function

### DIFF
--- a/src/api/controller/api/publish.js
+++ b/src/api/controller/api/publish.js
@@ -3,7 +3,6 @@ import getDocumentTransformer from '../../../common/getDocumentTransformer';
 
 export const publishMiddleware = async (ctx) => {
     const count = await ctx.dataset.count({});
-    let handled = 0;
 
     // TODO: remove this for stable state 2
     const firstLines = await ctx.dataset.findLimitFromSkip(1, 0);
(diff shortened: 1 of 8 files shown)


src/api/controller/api/publish.js
```diff
@@ -14,15 +13,32 @@ export const publishMiddleware = async (ctx) => {
...
-        const transformedDataset = await Promise.all(dataset.map(transformDocument));
+        const transformedDataset = await Promise.all(dataset.map(transformDocumentPreStage));
         await ctx.publishedDataset.insertMany(transformedDataset);
         handled += dataset.length;
     }
 
+    const transformDocumentPostStage = transformDocumentFactory('post');
+    handled = 0;
+
+    while (handled < count) {
```
ThieryMichel, line 34: Should'nt the uri be added at the start, and then th Link operation could be executed at the same time as all the other.

ThieryMichel (changes requested):

src/api/controller/api/publish.js
djhi: I'm not sure, can't it be the result of other transformed fields too ?

djhi: Which also means it shouldn't be added automatically as the first field too btw

djhi:
Superceded by #33

PR #32, opened 2017-02-02
[RFR] UI for fields required properties edition

djhi:
- [x] Added inputs for required properties on field form
- [x] Added an autocomplete for the scheme property

NOTE: There is a bug with the material-ui Autocomplete component which results in not being able to see all the found schemes when the menu is too close to the bottom of the page.

Display on public page will be addressed on another pr


ThieryMichel:
need rebase

PR #37, opened 2017-02-03
[RFR] Fix link publication

ThieryMichel:


src/api/controller/api/publish.js
```diff
@@ -1,20 +1,40 @@
...
+};
+
 export const publishMiddleware = async (ctx) => {
     const count = await ctx.dataset.count({});
-    let handled = 0;
 
     const columns = await ctx.field.findAll();
 
-    const transformDocument = ctx.getDocumentTransformer({ env: 'node', dataset: ctx.dataset }, columns);
+    const uriIndex = columns.findIndex(col => col.name === 'uri');
```
djhi: Can't you simply use `columns.find(col => col.name === 'uri')` here and pass the column found to `getDocumentTransformer` ?

src/api/controller/api/publish.js
```diff
@@ -1,20 +1,40 @@
...
-    }
+    await tranformAllDocument(
+        count,
+        ctx.dataset.findLimitFromSkip,
+        ctx.uriDataset.insertMany,
+        addUri,
+    );
+
+    const transformDocument = ctx
+        .getDocumentTransformer({ env: 'node', dataset: ctx.dataset }, columns.splice(uriIndex));
```
djhi: use `filter` instead of `splice`

src/api/models/uriDataset.js
```diff
@@ -0,0 +1,6 @@
+export default (db) => {
+    const collection = db.collection('dataset');
```
djhi: Why another model if targeting the same collection ?

djhi (changes requested):
Forgot to add comments as changes requests..

src/api/controller/api/publish.js
```diff
@@ -1,25 +1,82 @@
...
+import route from 'koa-route';
+
 /* eslint no-await-in-loop: off */
 import getDocumentTransformer from '../../../common/getDocumentTransformer';
 
-export const publishMiddleware = async (ctx) => {
-    const count = await ctx.dataset.count({});
+const app = new Koa();
+
+export const tranformAllDocument = async (count, findLimitFromSkip, insertBatch, transformer) => {
```
djhi: s/tranformAllDocument/tranformAllDocuments/

src/api/controller/api/publish.js
```diff
@@ -1,25 +1,82 @@
...
+    uri: doc.uri,
+});
+
+export const doPublish = async (ctx) => {
+    const count = await ctx.dataset.count({});
 
     const columns = await ctx.field.findAll();
 
-    const transformDocument = ctx.getDocumentTransformer({ env: 'node' }, columns);
+    const uriIndex = columns.findIndex(col => col.name === 'uri');
```
djhi: use `find`

src/api/controller/api/publish.js
```diff
@@ -1,25 +1,82 @@
...
+        ctx.dataset.findLimitFromSkip,
+        ctx.uriDataset.insertBatch,
+        addUri,
+    );
+
+    const transformDocument = ctx
+        .getDocumentTransformer({
+            env: 'node',
+            dataset: ctx.uriDataset,
+        }, columns.filter((_, index) => index !== uriIndex));
```
djhi: filter is good but you could filter on the name directly

src/api/controller/api/publish.js
```diff
@@ -1,25 +1,82 @@
...
+        ctx.uriDataset.findLimitFromSkip,
+        ctx.publishedDataset.insertBatch,
+        transformDocumentAndKeepUri,
+    );
 
     ctx.redirect('/api/publication');
 };
 
-export default async (ctx) => {
+export const tryPublish = async (ctx, next) => {
```
djhi: `tryPublish` ? I expected a transformation to occur here. Isn't it an initializer for the post middleware ?

djhi (changes requested):

src/api/controller/api/publish.js
ThieryMichel: it both initialize ctx, and catch publish error, maybe I should split it in two, and rename to preparePublish and handlePublishError.

djhi (approved):

PR #38, opened 2017-02-03
[RFR] Characteristics with dataset cover - public UI

djhi:
- [x] Introduce `publishedCharacteristics` collection
- [x] Updated `publish` feature to create the publishedCharacteristics entries
- [x] Added public UI components to display them on List page
- [x] e2e tests

screen

### DIFF
--- a/src/app/js/admin/fields/sagas/updateField.js
+++ b/src/app/js/admin/fields/sagas/updateField.js
@@ -30,8 +30,8 @@ export function* handleUpdateField({ meta: { form } }) {
         yield put(updateFieldError(error));
     } else {
         yield put(updateFieldSuccess(response));
+        yield put(updateField(fieldData));
     }
-    yield put(updateField(fieldData));
 }
 
 export default function* watchLoadField() {
--- a/src/app/js/admin/fields/sagas/updateField.js
+++ b/src/app/js/admin/fields/sagas/updateField.js
@@ -8,7 +8,7 @@ import {
 import {
     getFieldFormData,
     getUpdateFieldRequest,
-    updateField,
+    refreshField,
     updateFieldError,
     updateFieldSuccess,
 } from '../';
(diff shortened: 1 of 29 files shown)


src/app/js/admin/fields/sagas/updateField.js
```diff
@@ -30,8 +30,8 @@ export function* handleUpdateField({ meta: { form } }) {
         yield put(updateFieldError(error));
     } else {
         yield put(updateFieldSuccess(response));
+        yield put(updateField(fieldData));
```
ThieryMichel: updateField name seems wrong, we just uploaded the field successfully, and then we call updateField action?

src/app/js/admin/fields/sagas/updateField.js
```diff
@@ -30,8 +30,8 @@ export function* handleUpdateField({ meta: { form } }) {
         yield put(updateFieldError(error));
```
ThieryMichel, line 30: add a return to avoid the else

src/app/js/dataset/Dataset.js
```diff
@@ -39,7 +39,7 @@ export class DatasetComponent extends Component {
         return (
             <Card className="dataset">
                 <CardText>
-                    <Table selectable={false} fixedHeader={false} style={{ width: 'auto' }}>
+                    <Table selectable={false} fixedHeader={false} style={{ width: 'auto', minWidth: '100%' }}>
```
ThieryMichel: No literal declared in render to avoid useless rerendering.

ThieryMichel (changes requested):

src/app/js/admin/fields/sagas/updateField.js
djhi: Done

ThieryMichel (approved):

PR #45, opened 2017-02-07
[RFR] Field validation

djhi:


src/api/services/initializeFields.spec.js
```diff
@@ -27,6 +28,8 @@ describe('initializeFields', () => {
         await initializeFields(ctx, () => {});
 
         expect(ctx.field.insertOne).toHaveBeenCalledWith({
+            cover: COVER_COLLECTION,
+            label: URI_FIELD_NAME,
             name: URI_FIELD_NAME,
             cover: 'collection',
```
parmentf: Line 34 will overwrite cover field (from line 31).

djhi: wrong rebase. Good catch !

PR #53, opened 2017-02-09
[RFR] Remove resource

ThieryMichel:


package.json
```diff
@@ -47,6 +47,7 @@
     "lodash.omit": "4.5.0",
     "lodash.pickby": "4.6.0",
     "material-ui": "0.16.7",
+    "moment": "^2.17.1",
```
djhi: `^` is evil

src/api/controller/api/export.js
```diff
@@ -24,7 +24,7 @@ export async function exportMiddleware(ctx, type) {
         const exportStreamFactory = ctx.getExporter(type);
         const characteristics = await ctx.publishedCharacteristic.find({}).toArray();
         const fields = await ctx.field.find({}).toArray();
-        const publishedDatasetStream = await ctx.publishedDataset.find({}).stream();
+        const publishedDatasetStream = ctx.publishedDataset.getFindAllStream();
```
djhi, line 27: :+1:

src/api/models/publishedDataset.js
```diff
@@ -4,8 +4,18 @@ import omit from 'lodash.omit';
 
 export default (db) => {
     const collection = db.collection('publishedDataset');
+
     collection.insertBatch = documents => chunk(documents, 100).map(data => collection.insertMany(data));
-    collection.findLimitFromSkip = (limit, skip) => collection.find().skip(skip).limit(limit).toArray();
+
+    collection.findLimitFromSkip = (limit, skip) =>
```
djhi: would be better to add an optional `filter` parameter here and use this method in `findPage` imo

src/app/js/resource/Detail.js
```diff
@@ -12,20 +15,30 @@ import {
...
+                <CardActions>
+                    <Link to={{ pathname: '/resource/edit', query: { uri: resource.uri } }}>
+                        <FlatButton className="edit-resource" label={polyglot.t('edit')} primary />
+                    </Link>
+                    <Link to={{ pathname: '/resource/hide', query: { uri: resource.uri } }}>
+                        <FlatButton className="hide-resource" label={polyglot.t('hide')} primary />
+                    </Link>
+                </CardActions>
+            :
+                <span />
```
djhi: why not returning `null` ?

src/app/js/resource/RemovedDetail.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+RemovedDetailComponent.propTypes = {
+    reason: PropTypes.string,
+    removedAt: PropTypes.string,
+    p: polyglotPropTypes.isRequired,
+};
+
+const mapStateToProps = state => ({
+    ...getRemovedData(state),
```
djhi: return `getRemovedData(state)` directly

src/app/js/resource/index.spec.js
```diff
@@ -8,9 +8,12 @@ import reducer, {
     SAVE_RESOURCE,
     SAVE_RESOURCE_SUCCESS,
     SAVE_RESOURCE_ERROR,
+    HIDE_RESOURCE,
+    HIDE_RESOURCE_SUCCESS,
+    HIDE_RESOURCE_ERROR,
 } from './index';
 
-describe('resourceReducer', () => {
+describe.only('resourceReducer', () => {
```
djhi: only spotted

src/app/js/resource/routes.js
```diff
@@ -1,19 +1,51 @@
 import React from 'react';
 
 import Resource from './Resource';
+import { isLoggedIn as selectIsLoggedIn } from '../user';
+
+export const onEnterWithAuthenticationCheck = store => (nextState, replaceState) => {
```
djhi, line 6: We should extract this function from `admin/routes.js` and avoid duplication

djhi (changes requested):

src/app/js/resource/Detail.js
ThieryMichel: Last time I tried something like that it displayed null. Has it changed?

src/app/js/resource/routes.js
ThieryMichel: The logic is different, here i do not redirect toward login.

src/app/js/resource/RemovedDetail.js
djhi: Forgot this one ?

src/app/js/resource/routes.js
djhi: nvm, they really are too different.

djhi (approved):

PR #54, opened 2017-02-10
[RFR] Admin page to manage removed resources

djhi:
[Trello card](https://trello.com/c/J8411QYF)

PR #55, opened 2017-02-10
[RFR] Stabilize e2e tests again

djhi:
Follow #54

### DIFF
--- a/src/common/tests/conditions.js
+++ b/src/common/tests/conditions.js
@@ -12,7 +12,7 @@ export const inputElementIsFocusable = element =>
             element.getTagName(),
         ]);
 
-        if (tagName.toLowerCase() !== 'input' && tagName.toLowerCase() !== 'textarea') {
+        if (!['input', 'textarea'].includes(tagName.toLowerCase())) {
             throw new Error('Invalid element: must be an input or a textarea');
         }
 
--- a/src/common/tests/conditions.js
+++ b/src/common/tests/conditions.js
@@ -5,11 +5,16 @@ export const elementValueIs = (element, text) =>
         element.getAttribute('value').then(t => (t === text ? element : null)));
 
 export const inputElementIsFocusable = element =>
-    new WebElementCondition('until element is clicked', async () => {
-        const [isDisplayed, isEnabled] = await Promise.all([
+    new WebElementCondition('until element is focusable', async () => {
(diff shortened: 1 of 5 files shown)


src/common/tests/conditions.js
```diff
@@ -4,7 +4,32 @@ export const elementValueIs = (element, text) =>
...
+            element.isDisplayed(),
+            element.isEnabled(),
+        ]);
+
+        if (!isDisplayed || !isEnabled) return null;
+
+        return element
+                .sendKeys('')
+                .then(() => element)
+                .catch(() => null);
```
ThieryMichel, line 24: verify the error concern the focus.
We should keep the error if it says the element is not an input for example

src/common/tests/conditions.js
```diff
@@ -4,7 +4,32 @@ export const elementValueIs = (element, text) =>
...
+            element.isDisplayed(),
+            element.isEnabled(),
+        ]);
+
+        if (!isDisplayed || !isEnabled) return null;
+
+        return element
+                .click()
+                .then(() => element)
+                .catch(() => null);
```
ThieryMichel, line 39: verify the error

ThieryMichel (changes requested):

src/common/tests/conditions.js
djhi: I have no idea what to catch

djhi: I added a check for valid element. Can't do better

src/common/tests/conditions.js
djhi: As discussed, there is no correct and viable way to do that

djhi: Because selenium will throw a `WebDriverError` in all cases with only the message describing the exact prb. This message might change in future versions

ThieryMichel (approved):

src/common/tests/conditions.js
```diff
@@ -12,7 +12,7 @@ export const inputElementIsFocusable = element =>
             element.getTagName(),
         ]);
 
-        if (tagName.toLowerCase() !== 'input' && tagName.toLowerCase() !== 'textarea') {
+        if (!['input', 'textarea'].includes(tagName.toLowerCase())) {
```
ThieryMichel: :+1:

PR #60, opened 2017-02-13
[RFR] Formats

djhi:
[Trello card](https://trello.com/c/sxcxNlqd)

- [x] Add format property to field model
- [x] List format dynamically like transformers
- [x] Add display component based on format (type url)
- [x] Tests (to be continued)


src/app/js/dataset/DefaultColumn.js
```diff
@@ -0,0 +1,62 @@
...
+import fetchByUri from '../lib/fetchByUri';
+import { getToken } from '../user';
+import { field as fieldPropTypes } from '../propTypes';
+
+export class DatasetColumnComponent extends Component {
+    componentWillMount() {
+        const { column, resource } = this.props;
+        const linkTransformer = column.transformers && column.transformers.find(t => t.operation === 'LINK');
+        if (linkTransformer) {
+            const uri = resource[linkTransformer.args.find(a => a.name === 'reference').value];
```
ThieryMichel, line 18: it should not be the component that should make this call. But rather the saga that handle the change.

src/app/js/formats/uri/component.js
```diff
@@ -0,0 +1,37 @@
+import React, { PropTypes } from 'react';
+import { Link } from 'react-router';
+import { field as fieldPropTypes } from '../../propTypes';
+
+const UriComponent = ({ linkedResource, resource, field }) => {
```
ThieryMichel: i'll call it UriView, and don't forget the uppercase on the file name, all component start with one.

src/app/js/formats/uri/editionComponent.js
```diff
@@ -0,0 +1,77 @@
+import React, { Component, PropTypes } from 'react';
```
ThieryMichel: same for file name EditionComponent

ThieryMichel (changes requested):

src/app/js/dataset/DefaultColumn.js
djhi: what change ?

djhi: Imo, the only way to put it in saga is to (in `/dataset/sagas.js`)

- check for fields definitions and detect LINK
- loop over retrieved resources and call the fetchByUri for each linked column

djhi: Or to set up actions à la Twitter with normalized array of resources `{ data: { ... }, fetched: true|false }` etc. Seems overkill for now

ThieryMichel: No in updateField saga, when we change field transformer value, you get the change, and then check the uri.

djhi: we are on the public side here

ThieryMichel: my bad

ThieryMichel (approved):

PR #62, opened 2017-02-15
[RFR] Fix LINK function when reference is null

djhi:


src/api/controller/api/fetchLineBy.js
```diff
@@ -0,0 +1,8 @@
+export default ctx => (fieldName, value) =>
+    ctx.dataset
```
ThieryMichel: return null if value is null

src/api/controller/api/fetchLineBy.js
```diff
@@ -0,0 +1,8 @@
...
+        .findBy(fieldName, value)
+        .then(line => ({
+            ...line,
+            uri: `uri to ${fieldName}: ${value}`,
+        }))
+        .catch(() => null);
```
ThieryMichel: do not catch this error.

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #74, opened 2017-02-17
[RFR] Isolate sub domains by adding globalized selectors (Follow 72)

djhi:
Follow #72 

- [x] Add globalized selectors for each subdomain
- [x] Ensure with refer to the fields sub domain each time we needs fields
- [x] Publication sub domain responsability reduced to check whether data has been published


djhi:
Closed in favor of #77.

I messed up with my rebases ^^

PR #92, opened 2017-02-21
[WIP] Layout customization

djhi:
[Trello card](https://trello.com/c/fs7XflmL)


src/api/controller/front/dev.js
```diff
@@ -1,30 +1,46 @@
...
-        },
-    }));
+    app.use(async (ctx, next) => {
+        if (!customizationApplied) {
+            customizationApplied = true;
+
+            await copyCustomization();
+        }
+
+        const webpackConfig = require('../../../app/webpack.config.babel').default; // eslint-disable-line
```
djhi, line 18: Had to dynamically require the file, otherwise the calls to `path.resolve` inside it wouldn't succeed as the index.html is not yet copied

ThieryMichel: Why not have webpack.config.babel expose a function ?

djhi:
Superseded by #96

PR #96, opened 2017-02-22
[RFR] Layout customization

djhi:
- [x] Enable `index.html` customization
- [x] Split public and admin apps in webpack


ThieryMichel:
:+1:

PR #104, opened 2017-02-23
[RFR] Upload file extension

ThieryMichel:


README.md
```diff
@@ -65,23 +65,23 @@ import parseCsv from './parseCsv'; // eslint-disable-line
...
 };
 
 ```
 
 Notice how the key will determine the name of the loader.
-This name must match the content-type of the target file.
+This name must match the extension of the target file.
 This is how we determine which loader to use.
-Thus, a text/csv loader must be exported as text/csv.
+Thus, a loader for `.csv` file must be exported as csv.
```
djhi: s/exported as csv/exported as `csv`/

djhi (changes requested):
:+1:

djhi (approved):

PR #106, opened 2017-02-23
[RFR] Technical name

ThieryMichel:


djhi:
What about the transformers which refers columns ?

djhi (approved):

PR #112, opened 2017-02-28
new csv parser

touv:
try to add new loader...


package.json
```diff
@@ -41,6 +41,8 @@
     "copy-webpack-plugin": "4.0.1",
     "csv-parse": "1.1.10",
     "expect": "1.20.2",
+    "ezs": "^1.1.3",
```
djhi: Please fix the version: don't use `^` nor `~`

src/api/loaders/csv.js
```diff
@@ -0,0 +1,9 @@
+import ezs from 'ezs';
+import ezsBasics from 'ezs-basics';
+
+ezs.use(ezsBasics);
+
+export default config => stream =>
+  stream
+  .pipe(ezs('CSVParse'))
```
djhi: Just a note but indenting would make this easier to read

djhi (changes requested):

src/api/loaders/csv.js
```diff
@@ -0,0 +1,9 @@
+import ezs from 'ezs';
+import ezsBasics from 'ezs-basics';
+
+ezs.use(ezsBasics);
+
+export default config => stream =>
+  stream
```
djhi: Better but you should configure eslint for your IDE. This project has a specific eslint configuration with tabs made of 4 spaces :)

djhi (changes requested):
Can you prefix your PR's title with either `[WIP]` or `[RFR]` so that we know if/when you're done with it ?

src/api/loaders/csv.js
touv: ok, but where is the specific config file, I don't find any file like this https://github.com/Inist-CNRS/lodex-v1/blob/master/.eslintrc.js

touv: found !!! 
https://github.com/Inist-CNRS/lodex/blob/master/src/api/.eslintrc

djhi (approved):

PR #118, opened 2017-03-02
[RFR] Fix deconnexion

ThieryMichel:


src/api/controller/api/login.js
```diff
@@ -1,7 +1,7 @@
 import { auth } from 'config';
 import jwt from 'jsonwebtoken';
 
-export const loginMiddleware = (expDate = Date.now()) => (ctx) => {
+export default (ctx, expDate = Date.now()) => {
```
djhi: a middleware take ctx and next as parameters. You can't have a custom expDate arg like this

djhi (changes requested):

djhi (approved):

PR #121, opened 2017-03-06
[RFR] Language property

djhi:
[Trello card](https://trello.com/c/S8arIafZ)

screen

### DIFF
--- a/config.json
+++ b/config.json
@@ -21,6 +21,7 @@
        "csv"
    ],
    "languages": [
-       { "code": "fr", "label": "Français" }
+       { "code": "fr", "label": "Français" },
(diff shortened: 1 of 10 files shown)


config.json
```diff
@@ -19,5 +19,8 @@
     },
    "exporters": [
        "csv"
+   ],
+   "languages": [
+       { "code": "fr", "label": "Français" }
```
ThieryMichel: Add `en` as another language so that we have at least one choice

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #130, opened 2017-03-08
[RFR] Limit width of columns

djhi:
[Trello](https://trello.com/c/zd1rHQOR)

Also added missing texts for import saga

### DIFF
--- a/src/app/js/admin/parsing/ParsingExcerptColumn.js
+++ b/src/app/js/admin/parsing/ParsingExcerptColumn.js
@@ -1,0 +1,20 @@
+import React, { PropTypes } from 'react';
+import pure from 'recompose/pure';
+import { TableRowColumn } from 'material-ui/Table';
+import { isLongText, getShortText } from '../../lib/longTexts';
+
+export const ParsingExcerptColumnComponent = ({ value }) => (
+    isLongText(value)
+    ? (
+        <TableRowColumn title={value}>
+            {getShortText(value)}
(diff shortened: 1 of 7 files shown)


src/app/js/lib/longTexts.js
```diff
@@ -0,0 +1,3 @@
+export const isLongText = (text = '') => text && text.length > 50;
+
+export const getShortText = (text = '') => `${text.substr(0, 47)}...`;
```
ThieryMichel, line 3: merge the two functions shortening the text only if needed.

src/app/js/admin/parsing/ParsingExcerptColumn.js
```diff
@@ -0,0 +1,20 @@
+import React, { PropTypes } from 'react';
+import pure from 'recompose/pure';
+import { TableRowColumn } from 'material-ui/Table';
+import { isLongText, getShortText } from '../../lib/longTexts';
+
+export const ParsingExcerptColumnComponent = ({ value }) => (
+    isLongText(value)
```
ThieryMichel, line 7: move the isLongText logic inside getShortText

ThieryMichel (dismissed):

src/app/js/admin/parsing/ParsingExcerptColumn.js
djhi: No, I don't output the same component only if isLongText returns true

PR #140, opened 2017-03-09
[RFR] Facets

djhi:
[Trello card](https://trello.com/c/dagbcExp)

- [x] Added `isFacet` property to fields
- [x] Added and populated `publishedFacet` collection
- [x] Public UI for facets
- [x] Tests

### DIFF
--- a/src/api/controller/api/publish.js
+++ b/src/api/controller/api/publish.js
@@ -112,7 +112,6 @@ export const doPublish = async (ctx) => {
         transformDocumentAndKeepUri,
     );
     await ctx.publishCharacteristics(ctx, datasetCoverFields, count);
-    console.log({ facetFields });
     await ctx.publishFacets(ctx, facetFields);
 
     ctx.redirect('/api/publication');
--- a/src/api/controller/api/publish.js
+++ b/src/api/controller/api/publish.js
@@ -53,11 +53,19 @@ export const publishCharacteristics = async (ctx, datasetCoverFields, count) => 
     }
 };
 
+export const publishFacets = async (ctx, facetFields) =>
+    Promise.all(facetFields.map(field =>
+        ctx.publishedDataset
+            .findDistinctValuesForField(field.name)
+            .then(values => ctx.publishedFacet.insertFacet(field.name, values)),
+    ));
+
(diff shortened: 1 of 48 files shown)


src/api/controller/api/publish.js
```diff
@@ -103,6 +112,8 @@ export const doPublish = async (ctx) => {
         transformDocumentAndKeepUri,
     );
     await ctx.publishCharacteristics(ctx, datasetCoverFields, count);
+    console.log({ facetFields });
```
ThieryMichel: debug spotted

src/api/models/publishedFacet.js
```diff
@@ -0,0 +1,43 @@
...
+
+    collection.insertBatch = documents => chunk(documents, 100).map(data => collection.insertMany(data));
+
+    collection.insertFacet = (field, values) =>
+        collection.insertBatch(values.map(value => ({
+            field,
+            value,
+        })));
+
+    collection.findLimitFromSkip = (limit, skip, filter) =>
```
ThieryMichel, line 14: refactor to return both count and data
```js
const cursor = collection.find(...).sort(...)

return {
    count: await cursor.count(),
    data: await cursor.skip(...).limit(...).toArray(),
};
```

src/api/models/publishedFacet.js
```diff
@@ -0,0 +1,43 @@
...
+        collection
+            .find(filter)
+            .skip(skip)
+            .limit(limit)
+            .sort({ value: 1 })
```
ThieryMichel, line 19: sort before skip and limit otherwise you sort a subset of the data

ThieryMichel (changes requested):

src/api/models/publishedFacet.js
djhi: I'm quite sure that's not true as the query is only executed when yielding `toArray`

src/api/models/publishedFacet.js
djhi: Will do for all models in a separate PR

ThieryMichel (approved):

PR #146, opened 2017-03-13
[RFR] Fixes

### LINKED ISSUE #135: Setting a scheme not in the auto-completion list is not persisted
When I type `actor` in `scheme` auto-completed field, I get some suggestions:
- po:actor
- schema:actor

After searching the [LOV](http://lov.okfn.org/dataset/lov/terms?q=actor), I find http://id.loc.gov/vocabulary/relators/act which I find more appropriate.

I copy-paste this URI into the `scheme` field of the `actors` column, and `CLOSE` it.

Then, I click on the `actors` column in the preview, the panel open up again, and the `scheme` value is now  http://schema.org/actor, which corresponds to the schema:actor previous suggestion.

I lost the real value I wanted in this field: http://id.loc.gov/vocabulary/relators/act

djhi:
- [x] Fix React warnings on Dataset
- [x] Fix model export
- [x] Allow free values in scheme

Fixes #135


ThieryMichel:
:+1:

PR #151, opened 2017-03-14
[RFR] Fix public list pagination

djhi:


src/app/js/lib/Pagination.js
```diff
@@ -53,7 +53,8 @@ class Pagination extends Component {
 	static propTypes = {
 		onChange: PropTypes.func.isRequired,
 		total: PropTypes.number.isRequired,
-		perPage: PropTypes.number,
+		currentPage: PropTypes.number.isRequired,
```
ThieryMichel: What about replacing the tab with space?

src/app/js/lib/Pagination.js
```diff
@@ -152,7 +125,7 @@ class Pagination extends Component {
 						{
 							pages.map(page => (
 								<MenuItem 
-									primaryText={page}
+									primaryText={page + 1}
```
ThieryMichel: What if there is 10000 pages? (I had the case, it froze the browser). 
Replace with a text input.

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #154, opened 2017-03-14
[RFR] Selenium smart wait

ThieryMichel:
waiting for https://github.com/marmelab/selenium-smart-wait/pull/9

### DIFF
--- a/src/app/e2e/home_published.spec.js
+++ b/src/app/e2e/home_published.spec.js
@@ -1,10 +1,16 @@
 import { until, By } from 'selenium-webdriver';
 import expect from 'expect';
+import {
+    elementTextIs,
+    elementsCountIs,
+    elementIsClicked,
+    stalenessOf,
+} from 'selenium-smart-wait';
 
 import driver from '../../common/tests/chromeDriver';
 import { clear, loadFixtures } from '../../common/tests/fixtures';
 import fixtures from './home_published.json';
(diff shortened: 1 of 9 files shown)


src/app/e2e/home_published.spec.js
```diff
@@ -23,23 +29,17 @@ describe('Home page with published data', function homePublishedDataTests() {
         const properties = await driver.findElements(By.css('.dataset-characteristics .property'));
         expect(properties.length).toEqual(2);
 
-        const movieLabel = await driver.findElement(
-            By.css('.dataset-characteristics .property.movie .property_name'),
-        );
-        driver.wait(until.elementTextIs(movieLabel, 'Movie'), DEFAULT_WAIT_TIMEOUT);
+        const movieLabel = '.dataset-characteristics .property.movie .property_name';
```
djhi, line 32: I fail to see the value of declaring variables inside tests for the css selectors, except when they are used across several tests

djhi (changes requested):

src/app/e2e/home_published.spec.js
ThieryMichel: do you see that `.dataset-characteristics .property.movie .property_name` correspond to movieLabel ? me neither.

djhi: I do but as you like

djhi (approved):

PR #157, opened 2017-03-15
[RFR] Admin ergonomics - Publication

djhi:
> Do not release after merging this PR

[Trello card](https://trello.com/c/08ubfMng)

screen
screen
screen


src/app/e2e/admin/removedResources.spec.js
```diff
@@ -15,6 +15,11 @@ describe('Admin', () => {
         before(async () => {
             await clear(); // Had to ensure clear state for unknown reason
             await loadFixtures(fixtures);
+
+            await driver.get('http://localhost:3100/admin');
+            await driver.executeScript('return localStorage.clear();');
```
ThieryMichel, line 20: I don't think we put anything in localStorage

ThieryMichel (changes requested):

src/app/e2e/admin/removedResources.spec.js
djhi: Well, we clear it in all tests. Let's keep it that way

ThieryMichel (approved):

PR #169, opened 2017-03-17
[RFR] Save cancel field

ThieryMichel:
capture du 2017-03-20 16 22 02


djhi:
Looks good so far, can you add some screenshots too ?

PR #175, opened 2017-03-20
[RFR] Inist rdf exporters

touv:


touv:
@parmentf RDF export is available but it should be improved later with :
- grouped fields
- completed fields
- Dereferencing HTTP URIs
- and probably more !

djhi (dismissed):
Looks like you're missing an npm package: `n3`

docker-compose.yml
```diff
@@ -24,6 +24,7 @@ server:
         - mongo
     ports:
         - 3000:3000
+        - 8080:8080
```
djhi: why ?

package.json
```diff
@@ -43,14 +43,15 @@
...
-    "ezs-basics": "2.2.1",
+    "ezs-basics": "2.2.5",
     "file-loader": "0.9.0",
     "html-webpack-plugin": "2.26.0",
     "inist-ark": "2.0.0",
     "isomorphic-fetch": "2.2.1",
     "jbj": "7.1.4",
     "json-csv": "1.3.0",
     "json-loader": "0.5.4",
+    "jsonld": "^0.4.11",
```
djhi: please fix the versions by not using neither `^` nor `~`

djhi (dismissed):

package.json
djhi: You can configure your npm to always save exact versions: [article](http://c2fo.io/node/npm/javascript/2016/06/03/protecting-your-product-with-npm-save-exact/)

touv: ok, so we can use a per-project config file - /path/to/my/project/.npmrc 
is it a good idea ?

djhi (approved):

PR #180, opened 2017-03-21
[RFR] fix field scrollbar

ThieryMichel:


src/app/js/admin/publicationPreview/PublicationEditionModal.js
```diff
@@ -22,7 +22,7 @@ const styles = {
         marginRight: '1rem',
         paddingRight: '1rem',
         flexGrow: 1,
-        maxHeight: '50vw',
+        maxHeight: '55vh',
```
djhi, line 25: How does it fix the scrollbar in the modal ?

ThieryMichel: it's now based on viewport height instead of viewport width

djhi: doh

PR #188, opened 2017-03-22
[RFR] Version popover

ThieryMichel:
capture du 2017-03-22 16 08 08


djhi:
It should be placed on the far left

PR #196, opened 2017-03-23
[RFR] Fixes

djhi:
- [x] Fields should not be loaded by PublishButton but by PublicationPreview
- [x] Fix UPPERCASE transformer
- [x] Fix LINK transformer (getOtherLine) to avoid some useless requests
- [x] Attempt to stabilize e2e tests

### DIFF
--- a/src/common/transformers/getOtherLine.js
+++ b/src/common/transformers/getOtherLine.js
@@ -1,3 +1,8 @@
 export default (context, targetColumn) =>
-    async prev =>
-        context.fetchLineBy(targetColumn, prev);
+    (prev) => {
+        if (targetColumn) {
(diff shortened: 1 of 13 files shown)


src/common/transformers/getOtherLine.js
```diff
@@ -1,3 +1,8 @@
 export default (context, targetColumn) =>
-    async prev =>
-        context.fetchLineBy(targetColumn, prev);
+    (prev) => {
```
ThieryMichel: fetchLineBy return a promise aint it ?
In this case make this function async.
This way it will always return a promise, and it will be clear that it is async.

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #204, opened 2017-03-23
[RFR] allow to hide field once published

ThieryMichel:
capture du 2017-03-24 16 17 32

### DIFF
--- a/src/app/js/public/EditOntologyFieldForm.js
+++ b/src/app/js/public/EditOntologyFieldForm.js
@@ -52,11 +52,10 @@ EditOntologyFieldFormComponent.propTypes = {
     p: polyglotPropTypes.isRequired,
 };
 
-
 export default compose(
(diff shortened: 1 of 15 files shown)


src/app/js/public/EditOntologyFieldForm.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+
+export default compose(
+    withHandlers({
+        onSubmit: ({ onSaveProperty }) => (values) => {
+            onSaveProperty(values);
+        },
+    }),
+    withProps(({ field, ...props }) => ({
```
djhi: You don't need to pass own props with `withProps`

> Like mapProps(), except the newly created props are merged with the owner props.

djhi (changes requested):

djhi (approved):

PR #219, opened 2017-03-30
[RFR] Fix contribution moderation

djhi:
screen


ThieryMichel:
apply the style to the value too

PR #223, opened 2017-03-30
to avoid field names contains '+' and '/' characters

touv:
because base64 contains this characters https://en.wikipedia.org/wiki/Base64 and we cannot use them in web semantic export formats.

### DIFF
--- a/src/api/services/generateUid.js
+++ b/src/api/services/generateUid.js
@@ -6,6 +6,6 @@ export default () => new Promise((resolve, reject) =>
             reject(error);
             return;
         }
-        resolve(result.toString('base64'));
+        resolve(result.toString('base64').replace(/[+/]/g, 'z'));
     }),
 );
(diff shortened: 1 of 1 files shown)


src/api/services/generateUid.js
```diff
@@ -6,6 +6,6 @@ export default () => new Promise((resolve, reject) =>
             reject(error);
             return;
         }
-        resolve(result.toString('base64'));
+        resolve(result.toString('base64').replace(/[+/]/g, 'z'));
```
ThieryMichel: you need to escape the `/`

ThieryMichel (changes requested):

src/api/services/generateUid.js
touv: you're right, JS is so fun that this syntax works too ;-)

ThieryMichel (approved):

PR #224, opened 2017-03-30
[RFR] Punch lines

ThieryMichel:
capture du 2017-04-03 09 57 13
capture du 2017-04-03 09 57 23


djhi (changes requested):
- Padding is wrong in first screenshot: make it the same on each part.

- The punchline text of the URI dialog is not helpinig with this specific screen. It might be for the standard field but not for uri.

djhi (approved):

PR #225, opened 2017-04-01
[RFR] Improve export for complete fields

touv:


src/api/exporters/ezsLocals.js
```diff
@@ -84,7 +85,26 @@ export function JSONLDObject(data, feed) {
...
+            output[propertyName0] = {};
+            output[propertyName0][propertyName1] = data[propertyName1];
+            output[propertyName0][propertyName2] = data[propertyName2];
+            delete output[propertyName2];
+            output['@context'][propertyName0] = {};
+            output['@context'][propertyName0]['@id'] = output['@context'][propertyName2]['@id'];
+            output['@context'][propertyName2]['@id'] = 'http://www.w3.org/2000/01/rdf-schema#label';
+        };
+        const allFields = fields
+            .filter(field => field.completes !== undefined)
```
djhi: FYI: You don't need the `!== undefined` here. When simply returning `field.completes` it will be falsy if `null` or `undefined`

src/api/exporters/ezsLocals.js
```diff
@@ -84,7 +85,26 @@ export function JSONLDObject(data, feed) {
...
+            .map(field => transformCompleteFields(field));
```
djhi: You can pass `transformCompleteFields` directly to the map function like this:
```js
.map(transformCompleteFields);
```

src/api/exporters/ezsLocals.js
```diff
@@ -84,7 +85,26 @@ export function JSONLDObject(data, feed) {
                 }
             }
         });
-        feed.send(output);
+
+        const transformCompleteFields = async (field) => {
```
djhi: The function does not return anything. If you mutate, use `forEach` instead of map but you should avoid mutating parameters. I would keep a list of field without `completes` property and create a new array of fields with `completes` property, using the map function to get a transformed version of them. I would then combine the two lists.

djhi (dismissed):

src/api/exporters/ezsLocals.js
```diff
@@ -99,8 +99,8 @@ export async function JSONLDObject(data, feed) {
             output['@context'][propertyName2]['@id'] = 'http://www.w3.org/2000/01/rdf-schema#label';
         };
         const allFields = fields
-            .filter(field => field.completes !== undefined)
-            .map(field => transformCompleteFields(field));
+            .filter(field => field.completes)
+            .map(transformCompleteFields);
```
djhi: Aren't you only returning the fields with a completes property now ?

src/api/exporters/ezsLocals.js
djhi: You should avoid declaring functions inside other functions as you won't be able to test them.

src/api/exporters/ezsLocals.js
```diff
@@ -84,7 +92,26 @@ export function JSONLDObject(data, feed) {
                 }
             }
         });
-        feed.send(output);
+
+        const completesFields = fields
+            .filter(field => field.completes)
+            .map(transformCompleteFields);
+
+        Promise.all(completesFields).then((propertyNames) => {
```
ThieryMichel: const propertyNames = await Promise.all(completesField);

src/api/exporters/ezsLocals.js
```diff
@@ -49,13 +57,13 @@ export function linkDataset(data, feed) {
...
             '@id': scheme,
-//             '@type': 'https://www.w3.org/TR/xmlschema-2/#anyURI',
+            //             '@type': 'https://www.w3.org/TR/xmlschema-2/#anyURI',
         };
     }
     feed.send(data);
 }
 
-export function JSONLDObject(data, feed) {
+export async function JSONLDObject(data, feed) {
```
ThieryMichel: try and refactor this function without mutating a variable that has not be declared in the current scope.

src/api/exporters/ezsLocals.js
```diff
@@ -84,7 +92,26 @@ export function JSONLDObject(data, feed) {
...
+        const completesFields = fields
+            .filter(field => field.completes)
+            .map(transformCompleteFields);
+
+        Promise.all(completesFields).then((propertyNames) => {
+            propertyNames.forEach((field) => {
+                const propertyName0 = field[0];
+                const propertyName1 = field[1];
+                const propertyName2 = field[2];
+                output[propertyName0] = {};
```
ThieryMichel: do not mutate output here

src/api/exporters/ezsLocals.js
```diff
@@ -84,7 +92,26 @@ export function JSONLDObject(data, feed) {
...
             }
         });
-        feed.send(output);
+
+        const completesFields = fields
+            .filter(field => field.completes)
+            .map(transformCompleteFields);
+
+        Promise.all(completesFields).then((propertyNames) => {
+            propertyNames.forEach((field) => {
```
ThieryMichel: use .map

ThieryMichel (dismissed):

PR #229, opened 2017-04-03
[RFR] Rework URIs

djhi:
- [x] Use browserHistory to get clean URIs
- [x] Made baseUrl overridable through config.json

### DIFF
--- a/src/api/models/publishedDataset.js
+++ b/src/api/models/publishedDataset.js
@@ -2,6 +2,8 @@ import { ObjectID } from 'mongodb';
 import chunk from 'lodash.chunk';
 import omit from 'lodash.omit';
 import compose from 'lodash.compose';
+import config from 'config';
+import { getResourceUri } from '../../common/uris';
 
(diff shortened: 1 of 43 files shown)


src/api/models/publishedDataset.js
```diff
@@ -106,7 +108,14 @@ export default (db) => {
         )({ removedAt: { $exists: false } });
         const cursor = collection.getFindCursor(filters, sortBy, sortDir);
 
-        return cursor.stream();
+        return cursor
+            .map(resource => (
```
ThieryMichel, line 112: should we not do this during publication ?

ThieryMichel (changes requested):

src/api/models/publishedDataset.js
djhi: No because nothing prevent Julia to change the instance config after publication

ThieryMichel (approved):

PR #245, opened 2017-04-05
[RFR] Fix proxy

parmentf:
Maybe `.env` is not required (I did not tried without it, and I don't know how many caches there are, where they are, and how to reset them).

The proxy environment variables need to be taken into account at building.

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,0 @@
-http_proxy=http://proxyout.inist.fr:8080
-https_proxy=http://proxyout.inist.fr:8080
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+http_proxy=http://proxyout.inist.fr:8080
+https_proxy=http://proxyout.inist.fr:8080
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The only requirement is to provide an `index.html` file containing an element wi
 
 It is your responsability to include links to the administration and authentication pages:
 
-- administation: href must target `/admin`
+- administration: href must target `/admin`
 - authentication: href must target `#/login`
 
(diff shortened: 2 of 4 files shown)


parmentf:
I did not test either without the docker-compose v2 syntax (headers in `docker-compose.yml`.
I don't know if it's critical for the rest of docker process.

djhi (commented):
Where is `.env` used ? Is it automatically loaded by docker or docker-compose ?

parmentf:
Not sure `.env` is used. But it could be used by [docker-compose](https://docs.docker.com/compose/env-file/).

djhi:
Then it would be nice to document how to use it in the README. Maybe inside a new  *Troubleshooting* section ?

parmentf:
Yep. And I realize we added specific values in `.env`.

I should delete `lodex_server` image, rename `.env`, and launch `make docker-run-dev` again to see if docker-compose really needs `.env` or not.

parmentf:
OK. It's not useful.

djhi:
Forgot to mention it before but environment variables should be in uppercase. If you don't mind about this convention, just say so and I'll merge this pr

parmentf:
They ARE in lowercase in `docker-compose.yml` (and were before I went).

parmentf:
But I guess it would work in uppercase too: I define both cases in my bash.
In fact, [inist-tools](https://github.com/Inist-CNRS/inist-tools) do it for me.

parmentf:
I talked to @touv and we prefer to _add_ uppercase variables, without discarding lowercase ones.

That's because some softwares take into account one case, and others take only the other case.

PR #248, opened 2017-04-06
[RFR] Fix recursive completes

djhi:


config/test-dist.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
-    api_url: '',
-    port: 3000,
+    port: 3010,
```
djhi, line 2: Changed so that it is not the same as dev env by default

PR #249, opened 2017-04-06
[RFR] Format email and link

parmentf:
Adds `email` and `link` formats.

- `email`: adds a `mailto:` link in front page
- `link`: use the resource as a web link in front page


README.md
```diff
@@ -29,6 +29,14 @@ To access the mongo shell, run:
...
 ```
 
+### State reset
+
+During development, you may need to get an application state, where no data is published.
+
+One way is to modify `config/development.js`, and change the `dbName` to a new one.
+
+> **Note**: an alternative way is to use the mongo-shell and drop the database.
+
```
djhi: No, we supply two make commands for that:

`make clear-publication`: just clear the published data but keep your uploaded dataset and your model
`make clear-database`: clear the whole database

djhi (changes requested):

README.md
parmentf: Can I document these commands in the makefile, or is it a too detailed level?

djhi (approved):

ThieryMichel (changes requested):
The two adminComponent are identical, refactor them as one DefaultAdminComponentWithLabel

ThieryMichel (approved):

PR #251, opened 2017-04-07
[RFR] List format

djhi:
- [x] New SPLIT transformer
- [x] New LIST format

screen
screen

### DIFF
--- a/src/common/transformers/SPLIT.js
+++ b/src/common/transformers/SPLIT.js
@@ -1,0 +1,37 @@
+export const split = (value, separator) => {
+    if (!value) {
+        return [];
+    }
+
+    if (typeof value !== 'string') {
+        throw new Error('Invalid value: need a string');
+    }
+
+    return value.split(separator);
+};
+
+const transformation = (_, args) => (value) => {
+    const separator = args.find(a => a.name === 'separator');
+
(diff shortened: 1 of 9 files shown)


src/common/transformers/SPLIT.spec.js
```diff
@@ -0,0 +1,17 @@
...
+
+import { split } from './SPLIT';
+
+describe('SPLIT', () => {
+    it('should return splitted value', () => {
+        expect(split('hello dear world', ' dear ')).toEqual(['hello', 'world']);
+    });
+
+    it('should return empty string if value is null', () => {
+        expect(split(undefined, ' dear ')).toEqual('');
```
ThieryMichel: shouldn't it be an empty array ?

ThieryMichel (changes requested):

src/common/transformers/SPLIT.js
```diff
@@ -0,0 +1,37 @@
+export const split = (value, separator) => {
+    if (!value) {
+        return [];
```
ThieryMichel, line 3: Since you return an array here, how can you get an empty sting in test ?

ThieryMichel (approved):

PR #255, opened 2017-04-07
[RFR] Istex link

ThieryMichel:
capture du 2017-04-07 18 31 35


src/app/js/formats/istex/Component.js
```diff
@@ -0,0 +1,61 @@
...
+const styles = {
+    text: memoize(status => Object.assign({
+        fontSize: '1.5rem',
+        textDecoration: status === REJECTED ? 'line-through' : 'none',
+    })),
+};
+
+export const IstexView = ({ fieldStatus, data, error, field, resource, p: { t } }) => (
+    <span style={styles.text(fieldStatus)}>
+        <span>ISTEX results for {resource[field.name]}:</span>
```
djhi: useless line imo but if you do add it then make it translatable

src/app/js/formats/istex/Component.js
```diff
@@ -0,0 +1,61 @@
...
+    className: null,
+    fieldStatus: null,
+    shrink: false,
+    data: null,
+    error: null,
+};
+
+export default compose(
+    translate,
+    FetchPaginatedDataHOC(fetchIstexData),
```
djhi: s/FetchPaginatedDataHOC/fetchPaginatedData/

camelcase for functions and we don't need the suffix

src/app/js/formats/istex/IstexItem.js
```diff
@@ -0,0 +1,31 @@
...
+    abstract: PropTypes.string,
+};
+
+IstexItemComponent.defaultProps = {
+    abstract: null,
+};
+
+export default compose(
+    withHandlers({
+        handleOpenUrl: ({ fulltext }) => () => window.open(fulltext),
```
djhi: `window.open` is not handled the same way in all browsers and should be avoided. Try to use an anchor link (maybe with a target set to blank, however this is a bad practice too)

src/app/js/formats/istex/Component.js
```diff
@@ -0,0 +1,61 @@
...
+import IstexItem from './IstexItem';
+
+const styles = {
+    text: memoize(status => Object.assign({
+        fontSize: '1.5rem',
+        textDecoration: status === REJECTED ? 'line-through' : 'none',
+    })),
+};
+
+export const IstexView = ({ fieldStatus, data, error, field, resource, p: { t } }) => (
```
djhi: we use `p: polyglot` everywhere and we should keep it consistant

djhi (changes requested):

src/app/js/formats/istex/Component.js
```diff
@@ -4,7 +4,7 @@ import { List } from 'material-ui/List';
 import translate from 'redux-polyglot/translate';
 import compose from 'recompose/compose';
 
-import FetchPaginatedDataHOC from '../../lib/FetchPaginatedDataHOC';
+import fetchPaginatedDataForComponent from '../../lib/fetchPaginatedDataForComponent';
```
djhi: why the suffix ? In React, an HOC is necessarily for a component.

djhi (changes requested):

src/app/js/formats/istex/Component.js
djhi: s/fetchPaginatedDataForComponent/fetchPaginatedData/

ThieryMichel: Because it is in lib and I want to differentiate it from other function or component

ThieryMichel: and if I call it fetchPaginatedData it would seem to be a service to fetch paginated data, which it is not.

PR #257, opened 2017-04-10
[RFR] New JBJ transformer

djhi:
[Trello card](https://trello.com/c/HanspZVl)
screen


src/api/services/applyJbjStylesheet.js
```diff
@@ -0,0 +1,11 @@
+import { render } from 'jbj';
+
+export default (value, stylesheet) => new Promise((resolve, reject) => {
```
parmentf: I don't see any declarations of JBJ's module(s), so this renderer would be very limited.

Maybe [use](https://github.com/Inist-CNRS/node-jbj#usemodule-function--none) all [available modules](https://github.com/Inist-CNRS/node-jbj#actions) (only `basics` and `ejs` are present by default):
- `parse`
- `template`
- `array`
- `rdfa` (not sure it's useful there)
- `nlp`
- `numerical`
- `jsonld` (not sure it's useful there)

parmentf (approved):

PR #258, opened 2017-04-10
[RFR] Fix exports

djhi:


src/api/controller/api/export.js
```diff
@@ -79,7 +79,7 @@ export async function setup(ctx, next) {
 
 export async function exportMiddleware(ctx, type) {
     try {
-        const exporterConfig = ctx.getExporterConfig();
+        const exporterConfig = getExporterConfig();
```
ThieryMichel: inject it in ctx in setup instead

ThieryMichel (changes requested):

src/api/controller/api/export.js
djhi: why ?

ThieryMichel (approved):

PR #285, opened 2017-04-20
[RFR] Refactored lib folder and applied eslint fixes

djhi:


src/app/e2e/admin/publication.spec.js
```diff
@@ -73,7 +73,7 @@ describe('Admin', () => {
                 await waitForPreviewComputing();
 
                 for (let index = 1; index < 5; index += 1) {
```
ThieryMichel, line 75: replace with promises = `_.range(...).map(index => elementTextMatches(...)).map(condition => driver.wait(...))`

ThieryMichel (dismissed):

src/app/e2e/admin/publication.spec.js
djhi: no, it doesn't work

PR #286, opened 2017-04-20
[RFR] Extracted field edition inputs into distinct components

djhi:
+ fixed some react errors regarding currently edited field after save


src/app/js/lib/components/FieldDisplayInListInput.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
ThieryMichel, line 1: over splitting ? All the component you added could be rewritten as one HOC passing the component to the component property of Field.
```js
Component => ({ label, name, fullWidth}) => (
    <Field component={Component} label={label} ...  />
);
```
And then FormCheckBoxField, FormatEdition, etc... could use the HOC directly in their export.

ThieryMichel (changes requested):

src/app/js/lib/components/FieldDisplayInResourceInput.js
```diff
@@ -1,20 +1,10 @@
...
-    />
-);
-
-FieldDisplayInResourceInputComponent.propTypes = {
-    p: polyglotPropTypes.isRequired,
-};
-
-export default translate(FieldDisplayInResourceInputComponent);
+export default () => <FieldInput
+    name="display_in_resource"
```
ThieryMichel: the only difference with the other component are int the props, do not create a component just for that

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #308, opened 2017-04-29
[RFR] Add route for reducers

touv:


src/api/controller/api/reduce.js
```diff
@@ -0,0 +1,46 @@
...
+        ctx.body = {
+            data: [],
+            total: 0,
+        };
+    }
+};
+
+const app = new Koa();
+
+app.use(route.get('/:reducer', mapAndReduce));
```
djhi, line 44: I suggest to add a `/reducers/` prefix here to avoid conflicts: `/reducers/:reduxer`

djhi (changes requested):

src/api/controller/api/reduce.js
djhi: nvm, haven't noticed the mount...

djhi (approved):

PR #315, opened 2017-05-03
[RPR] Display an translation for all formats

touv:


src/app/js/i18n/translations/fr.js
```diff
@@ -224,4 +224,12 @@ export default {
...
     invalid_url: 'url is invalid',
     or: 'or',
+    html: 'Pour des balises HTML',
+    list: 'Pour une liste de valeurs',
+    uri: 'Pour un lien interne',
+    email: 'Pour un lien courriel',
+    image: 'Pour l\'URL d\'une image',
+    link: 'Pour une URL externe',
+    chart: 'Pour un graphique contextuel',
+    istex: 'Pour une requête ISTEX ',
```
parmentf: Would'nt it be better with `Balises HTML`, `URL externe`: shorter, better fit with English version?

parmentf (dismissed):
Remove `Pour ...`, to better fit English version.

src/app/js/i18n/translations/fr.js
touv: fixed

PR #318, opened 2017-05-03
[CANCELED] a try for issue#317

touv:


parmentf:
See https://github.com/Inist-CNRS/lodex/pull/329 (which fixed it another way).

PR #326, opened 2017-05-05
[RFR] Add Markdown format

touv:


src/app/js/i18n/translations/fr.js
```diff
@@ -232,4 +232,5 @@ export default {
     link: 'URL externe',
     chart: 'Graphique contextuel',
     istex: 'Requête ISTEX',
```
parmentf, line 234: Add the `markdown` key to the `en.js` translation file.

src/app/js/i18n/translations/fr.js
```diff
@@ -232,4 +232,5 @@ export default {
...
+    markdown: 'Syntax Markdown',
```
parmentf: In French, it's Syntax**e** Markdown.

Issue #321, opened 2017-05-04
Make lodex cleanup database at deletion in ezmaster


parmentf:
See https://github.com/Inist-CNRS/ezmaster/commit/f563abb85eeeb0644fc3c621fe151782516fe583

Because after deleting many instances, their mongo databases stay.
Two consequences:
1. When you create a new instance with the same name, the database already exists and has data.
2. After loading and deleting many instances, the disk space is not cleaned up.

parmentf:
3. When you create a new instance with the name of a deleted instance, which was published, the new instance is *already published!*

parmentf:
Add a `cleanupScript` key to [Dockerfile](https://github.com/Inist-CNRS/lodex/blob/master/Dockerfile), pointing to `/app/bin/clearDatabase.sh` in `bin`.

See https://github.com/Inist-CNRS/lodex/blob/master/Dockerfile (no docker-compose, replace `lodex` with the instance name).

Maybe written in node.

parmentf:
Done in [7.12.0](https://github.com/Inist-CNRS/lodex/releases/tag/v7.12.0).

PR #333, opened 2017-05-10
[RFR] Add a cleanup script

parmentf:
Meant to remove the mongo database at ezmaster's instance deleting.

See  #321.


parmentf:
It works now, but I had to relaunch ezmaster, because of the rebuilding of lodex's image (I assume that).

PR #341, opened 2017-05-16
[RFR] choose the number of topfields in the resource page

touv:
image


src/app/js/public/resource/Detail.js
```diff
@@ -116,8 +116,8 @@ export const DetailComponent = ({
     sharingUri,
     backToListLabel,
 }) => {
-    const topFields = fields.slice(0, 2);
-    const otherFields = fields.slice(2);
+    const topFields = fields.slice(0, Number(topFieldsCount));
```
parmentf: And what about `config.json` where `topFieldsCount` is not present?
We should add a default value.

parmentf (dismissed):
Add a default value for `topFieldsCount`.

PR #348, opened 2017-05-23
[WIP] number instance & error Proxy/API

QuanticPotatoes:
- Remove instance's number in URI.
- Catch error from proxy & API while execute scroll request.


src/api/exporters/ezsLocals/JSONLDObject.js
```diff
@@ -61,15 +61,28 @@ function mergeSimpleField(output, field, data) {
...
 
     return uri;
 }
 
+function getUri(uri) {
+    const u = removeNumberInstance(uri);
+
+    if (u.indexOf('http://') !== 0 &&
+        u.indexOf('https://') !== 0) {
+        return path.normalize(this.getParam('hostname', 'http://lod.istex.fr/').concat(u));
```
touv: lod.istex.fr should be define in the config file isn't it ?

parmentf: It seems that `this` is undefined within that function.

parmentf: @touv Why do you want to append `http://lod.istex.fr/` to any URI?
It should be what is called `host` (`http://wos-category.lod.istex.fr/`) in [default.js](https://github.com/Inist-CNRS/lodex/blob/master/config/default.js#L10), shouldn't it?

Besides, in what cases `u` does not begin with `http`?
default.js set a default value to `host`, beginning with `http`.

touv: now, "this" is undefined, before "this" was the ezs scope.  this.getParam('hostname') is used to fetch ezs statement parameter (https://github.com/touv/node-ezs#scope)

of course, we should use "host" from defaut.js, and to get it, one way is to use ezs parameter like this https://github.com/Inist-CNRS/lodex/blob/master/src/api/exporters/exportJsonld.js#L14

Issue #336, opened 2017-05-11
After a while running, no connection possible


parmentf:
When lodex has been running for a while (say, a day), you can't login except if you stop and start lodex again.

After typing your login and password, and clicking on `SIGN IN`, the user is redirected to the same page.

parmentf:
Weird: just after a successful login on front, exporting the model (from the `ONTOLOGY` tab) redirects to http://inist-category.lod.istex.fr/login (v7.12.0, same instance)

parmentf:
When you change `expiresIn` in [development-dist.js](https://github.com/Inist-CNRS/lodex/blob/master/config/development-dist.js#L7), from `10 * 3600` (ten hour) to `60` (one minute), you can reproduce the bug after very fewer time than a day.

QuanticPotatoes:
This issue was fixed.

PR #350, opened 2017-05-24
[RFR] Try to fix bug #336

touv:


parmentf:
Good fix. Now we'll have to fix the test. 😕

parmentf:
Fix #336 
Grand merci :)

PR #352, opened 2017-05-29
[RFR] Improve ontology tab

touv:
image


src/app/js/i18n/translations/en.js
```diff
@@ -249,4 +249,6 @@ export default {
     list_format_select_level: 'Choose title level',
     dataset: 'Dataset',
     paragraph: 'Paragraph',
+    identifier: 'Identifier',
+    count_of_field: '# doccuments',
```
parmentf: `doccuments` => `documents` (one superfluous character)

PR #358, opened 2017-06-04
[WIP] Add format lodexlink

touv:
to link 2 resources of 2 différents lodex easily

### DIFF
--- a/src/api/exporters/exportMin.js
+++ b/src/api/exporters/exportMin.js
@@ -1,0 +1,20 @@
+import ezs from 'ezs';
+import ezsBasics from 'ezs-basics';
+import ezsLocals from './ezsLocals';
+
+ezs.use(ezsBasics);
+ezs.use(ezsLocals);
+
+const exporter = (config, fields, characteristics, stream) =>
+    stream
+        .pipe(ezs('filterVersions'))
+        .pipe(ezs('filterContributions', { fields }))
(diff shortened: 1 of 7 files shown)


parmentf:
This feature might be linked to the one in 083e61878f96b6a50605ef3a0c5f3b0829d5e8cf : we can chose what field is the title, and what field is the description (for the Overview tab).

touv:
You 're right, in the future, default title (and description) should be define with the same way for : 
 - Minimal export / LODEX Resource Format 
 - Overview tab 
 - Back to list text label cf. https://github.com/Inist-CNRS/lodex/blob/619c60854fd599d4a0cb4284485587f88e4c7ece/src/app/js/fields/index.js#L450
 - Internal URI Format cf. https://github.com/Inist-CNRS/lodex/blob/619c60854fd599d4a0cb4284485587f88e4c7ece/src/app/js/formats/uri/Component.js#L25

parmentf:
Lacking tests?
These ones are quick to execute, they're not selenium test ;)
See 8136dcba4a4008934f908755b8f8e56654ed7213

touv:
yes, of course but i think it's a bit more complex. If we want to have a unit test for this component, we need to mock the http call. The goal of this component is to transform a LODEX URL to a link with url and label.  For me, it's hard and long to code with a small benefecit. It's just a format, if it's doesn't work, we can disable it.

PR #375, opened 2017-06-14
[RFR] Add default transformer

touv:
cqfd

### DIFF
--- a/src/common/transformers/DEFAULT.js
+++ b/src/common/transformers/DEFAULT.js
@@ -1,0 +1,32 @@
+export const defval = (value, alternative) => {
+    if (!value) {
+        return alternative;
+    }
+    return value;
+};
+
+const transformation = (_, args) => (value) => {
+    const alternative = args.find(a => a.name === 'alternative');
+
+    if (!alternative) {
+        throw new Error('Invalid Argument for DEFAULT transformation');
+    }
+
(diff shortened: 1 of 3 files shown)


src/common/transformers/DEFAULT.js
```diff
@@ -0,0 +1,32 @@
...
+            reject(error);
+        }
+    });
+};
+
+transformation.getMetas = () => ({
+    name: 'DEFAULT',
+    type: 'transform',
+    args: [
+        { name: 'alternative', type: 'string' },
```
parmentf, line 28: It it always a string? Why not a number?

src/common/transformers/DEFAULT.js
```diff
@@ -0,0 +1,32 @@
+export const defval = (value, alternative) => {
+    if (!value) {
```
parmentf, line 2: If value is a string, OK.
But if it can be a number, one should compare with `undefined` (but not sure what is given when no value is defined).

src/common/transformers/DEFAULT.js
touv: At the import, all fields are "string" so if we want "number", we can add a CAST transformer. And when we can chain DEFAULT and NUMBER to generate a default "number" value

PR #376, opened 2017-06-14
[RFR] same value filter in tabs

QuanticPotatoes:


src/app/js/public/dataset/Filter.js
```diff
@@ -24,36 +24,43 @@ const styles = {
...
-            <TextField
-                className="filter"
-                hintText={polyglot.t('filter')}
-                onChange={(_, e) => handleFilterChange(e)}
-                style={styles.textbox}
-            />
-        </ToolbarGroup>
-    )
-    : null);
+// export class FilterComponent extends Component {
```
parmentf, line 27: We should not let commented out code.
Doesn't eslint scream?

src/app/js/public/overview/Overview.js
```diff
@@ -4,7 +4,6 @@ import { connect } from 'react-redux';
 import compose from 'recompose/compose';
 import translate from 'redux-polyglot/translate';
 import { Card, CardHeader, Avatar } from 'material-ui';
-import LocalLibrary from 'material-ui/svg-icons/maps/local-library';
```
parmentf, line 7: Why remove `LocalLibrary`? No more used?

PR #392, opened 2017-06-26
[RFR] Add transformers 3

touv:
new transformers that enable to type/cast field value

### DIFF
--- a/src/common/transformers/BOOLEAN.js
+++ b/src/common/transformers/BOOLEAN.js
@@ -1,0 +1,40 @@
+export const toBoolean = (value) => {
+    if (!value) {
+        return false;
+    }
+
+    if (Array.isArray(value)) {
+        return value.map(toBoolean);
+    }
+
+    if (typeof value !== 'string') {
+        const val = value.trim();
+        if (val === '1' || value === 'true' || value === 'on') {
+            return true;
+        }
+        return false;
+    }
+
+    if (value === 1 || value === true) {
+        return true;
+    }
+
+    return false;
+};
+
+const transformation = () => value =>
+    new Promise((resolve, reject) => {
+        try {
+            resolve(toBoolean(value));
+        } catch (error) {
+            reject(error);
+        }
+    });
+
+transformation.getMetas = () => ({
+    name: 'BOOLEAN',
+    type: 'transform',
+    args: [],
+});
+
+export default transformation;
--- a/src/common/transformers/BOOLEAN.js
+++ b/src/common/transformers/BOOLEAN.js
@@ -7,7 +7,7 @@ export const toBoolean = (value) => {
         return value.map(toBoolean);
     }
 
-    if (typeof value !== 'string') {
+    if (typeof value === 'string') {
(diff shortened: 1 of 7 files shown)


src/common/transformers/BOOLEAN.js
```diff
@@ -0,0 +1,40 @@
...
+
+    if (typeof value === 'string') {
+        const val = value.trim();
+        if (val === '1' || value === 'true' || value === 'on') {
+            return true;
+        }
+        return false;
+    }
+
+    if (value === 1 || value === true) {
```
parmentf, line 18: Usually, the logic is the inverse: when the value is not falsy (0, false, ...) it is true.
Why not do that way?

src/common/transformers/BOOLEAN.spec.js
```diff
@@ -0,0 +1,20 @@
+import expect from 'expect';
+
+import { toBoolean } from './BOOLEAN';
+
+describe('BOOLEAN', () => {
+    it('should return true', () => {
+        expect(toBoolean('true')).toBe(true);
+    });
+
+    it('should return uppercase value of each item in array', () => {
```
parmentf: That test label seems to come from a copy/paste operation.

src/common/transformers/BOOLEAN.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        expect(toBoolean(['1', '0'])).toEqual([true, false]);
+    });
+
+    it('should return null if value is literal', () => {
```
parmentf: That one too...

src/common/transformers/NUMBER.spec.js
```diff
@@ -0,0 +1,20 @@
+import expect from 'expect';
+
+import { toNumber } from './NUMBER';
+
+describe('NUMBER', () => {
+    it('should return 123', () => {
+        expect(toNumber('1234')).toBe(1234);
+    });
+
+    it('should return uppercase value of each item in array', () => {
```
parmentf: Again.

src/common/transformers/STRING.spec.js
```diff
@@ -0,0 +1,20 @@
+import expect from 'expect';
+
+import { toString } from './STRING';
+
+describe('STRING', () => {
+    it('should return abcs', () => {
+        expect(toString(' hello ')).toBe('hello');
+    });
+
+    it('should return uppercase value of each item in array', () => {
```
parmentf: And again...

src/common/transformers/STRING.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        expect(toString([1, true])).toEqual(['1', 'true']);
+    });
+
+    it('should return null if value is literal', () => {
```
parmentf: And again.

src/common/transformers/BOOLEAN.js
parmentf: This should almost never happen (except if casted just before, the input is a string).

PR #411, opened 2017-07-17
[RFR] url not encoded in overview

QuanticPotatoes:


touv (commented):
overview tab & detail tab should be generate the same link, so we will need to upgrade this function https://github.com/Inist-CNRS/lodex/blob/master/src/app/js/public/dataset/UriColumn.js#L9-L15

touv:
immpove 
```
 const identifier = data.uri.substring(0, 5);		 +                        style={styles.item}
 const resource = data.uri.substring(5, data.uri.length);		
 const uri = validUrl.isWebUri(resource) ? `/${identifier}${encodeURIComponent(resource)}` : `/${data.uri}`;		
 ```

PR #416, opened 2017-07-27
[RFR ] Improve & add transformers

touv:
improve :  SPLIT & BOOLEAN 
add : TRUNCATE & SHIFT

### DIFF
--- a/src/common/transformers/SHIFT.js
+++ b/src/common/transformers/SHIFT.js
@@ -1,0 +1,35 @@
+export const shift = (value, arg) => {
+    if (!value) {
+        return value;
+    }
+    const gap = Number(arg);
+    return Array.isArray(value) || typeof value === 'string' ?
+        value.slice(gap) :
+        gap - value;
+};
+
+const transformation = (_, args) => (value) => {
+    const arg = args.find(a => a.name === 'gap');
+
+    if (!arg) {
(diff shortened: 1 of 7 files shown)


src/common/transformers/SHIFT.js
```diff
@@ -0,0 +1,35 @@
+export const shift = (value, arg) => {
+    if (!value) {
+        return value;
+    }
+    const gap = Number(arg);
+    return Array.isArray(value) || typeof value === 'string' ?
+        value.slice(gap) :
+        gap - value;
```
parmentf, line 8: Oh? Used to substract numbers?

src/common/transformers/TRUNCATE.spec.js
```diff
@@ -0,0 +1,13 @@
+import expect from 'expect';
+
+import { truncate } from './TRUNCATE';
+
+describe('TRUNCATE', () => {
+    it('#1', () => {
+        expect(truncate('hello world', 5)).toEqual('hello');
```
parmentf, line 7: So, the argument is the index (in the string/array) from where the values are dropped?

It could have been the number of values to drop (from the last one).

parmentf:
We will end up with a replicate of JBJ, after adding more and more complex transformers :wink:

PR #427, opened 2017-08-28
[WIP] fix field order

### LINKED ISSUE #419: Field order in published data is broken
When using the `Position` in the `How and where it is displayed` of the backoffice, no fields should have twice the same `position` index.

Treat the `after` position as an `insertBefore` in a singly linked list.

Ex:

```
Init:           A <-- after-- B <---after---- C

               positions: A:1, B:2, C:3

Action:      put D after B

Result:      A <--after--- B <---after---- D <---after--- C

               positions: A: 1, B: 2, C: 4, D: 3
```

QuanticPotatoes:
* add drag and drop in ontology
* fix #419


parmentf:
Done in #439

PR #455, opened 2017-10-18
[RFR] badge & ribbon for number

touv:
image
image


src/api/routines/count-all.ini
```diff
@@ -0,0 +1,28 @@
+# Reduce to ditinct values form one or more fields
```
parmentf: This is still confusing. ;P
Moreover, it's not the right comment.

parmentf:
Isn't this PR conflicting with #452's aim (make the application style-neutral)?

PR #456, opened 2017-10-19
[RFR] more neutral look for emphased numbers

touv:
image


parmentf:
In previous version of ezvis, administrators were used to icons, beside numbers.

PR #459, opened 2017-10-20
[RFR] Fix NQuads export with classes #443

### LINKED ISSUE #443: Export is broken when there are classes in the model
When you try to export an item of a dataset, which has classes, the loading does not end, and you get this error serverside:

```
 2017-10-20T14:01:01.158Z - info: /api/export/nquads?uri=uid%3A%2FNZ33VJ02 method=GET, remoteIP=::ffff:172.16.101.52, userAgent=Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.62 Safari/537.36, status=200
 You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
 TypeError: Cannot read property 'startsWith' of undefined
     at getUri (/app/src/api/statements/JSONLDObject/getUri.js:19:10)
     at mergeClasses (/app/src/api/statements/JSONLDObject/mergeClasses.js:10:19)
     at merge (/app/src/api/statements/JSONLDObject/index.js:21:32)
     at currentOutputPromise.then.currentOutput (/app/src/api/statements/JSONLDObject/index.js:63:21)
     at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

parmentf:
Fix #443

### DIFF
--- a/src/api/exporters/exportNQuads.spec.js
+++ b/src/api/exporters/exportNQuads.spec.js
@@ -5,27 +5,10 @@ import exportNQuads from './exportNQuads';
 
 const fields = [{
     cover: 'collection',
-    label: 'title',
-    transformers: [
-        {
-            operation: 'COLUMN',
-            args: [
-                {
-                    name: 'column',
-                    type: 'column',
-                    value: 'title',
-                },
-            ],
-        },
-    ],
     scheme: 'http://purl.org/dc/terms/title',
     format: {
         name: 'None',
     },
-    display_in_list: true,
-    display_in_resource: true,
-    searchable: true,
-    position: 3,
     name: 'Q98n',
 }, {
     cover: 'collection',
@@ -37,7 +20,7 @@ const fields = [{
     ],
 }];
 
-describe.only('export Nquads', () => {
+describe('export Nquads', () => {
     it('should export a single data property', (done) => {
         let outputString = '';
         exportNQuads(
--- a/src/api/exporters/exportNQuads.spec.js
+++ b/src/api/exporters/exportNQuads.spec.js
@@ -85,4 +85,59 @@ describe('export Nquads', () => {
             return feed.end();
         }));
     });
+
+    it('should export a composed object property (with a class)', (done) => {
+        let outputString = '';
+        exportNQuads(
+            {
+                cleanHost: '',
+                schemeForDatasetLink: '',
+            },
+            [{
+                cover: 'collection',
+                scheme: 'http://property/composed',
+                name: 'propcomposed',
+                classes: [
+                    'http://class/composed',
+                ],
+                composedOf: {
+                    fields: [
+                        'propb',
+                        'propc',
+                    ],
+                },
+            }, {
+                cover: 'collection',
+                scheme: 'http://property/b',
+                name: 'propb',
+            }, {
+                cover: 'collection',
+                scheme: 'http://property/c',
+                name: 'propc',
+            }],
+            null,
+            from([{
+                uri: 'http://uri/1',
+                propcomposed: 'label a',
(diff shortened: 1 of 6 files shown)


src/api/statements/JSONLDObject/mergeClasses.js
```diff
@@ -1,3 +1,4 @@
+import { host } from 'config';
```
parmentf, line 1: Oops. This should not be done in `statements`, as they are aimed to be exported to another npm package: `ezs-lodex`.

src/api/exporters/exportNQuads.spec.js
```diff
@@ -82,27 +63,30 @@ describe('export Nquads', () => {
...
-                ZI8w: 'A New Hope' }]),
+                uri: 'http://data.istex.fr/1',
+                D0n9: 'https://publication-type.data.istex.fr/ark:/67375/JMC-0GLKJH51-B' }]),
         ).pipe(ezs((data, feed) => {
             if (data !== null) {
                 outputString += data;
             } else {
                 try {
-                    expect(outputString).toEqual('<http://data.istex.fr> <http://purl.org/dc/terms/title> "A New Hope" .\n');
+                    expect(outputString).toEqual(`<http://data.istex.fr/1/classes/vlM0> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://data.istex.fr/ontology/istex#PublicationTypeConcept> .
```
parmentf: This expected output is a proof that the exemple is not yet the best.
A link should not have classes.
The pointed object should declare itself its rdf:type.

parmentf: Morever, the value of the property (`https://publication-type.data.istex.fr/ark:/67375/JMC-0GLKJH51-B`) is not exported in the triples!

parmentf:
It seems that sometimes, `classes` means the classes of the *domain* of a property (scheme), and sometimes, it means the classes of the *range* of a property!

```
domain object ------- scheme / property -------> range object
```

After discussion with a semantic web consultant, it seems obvious now that we only have to express **range** classes of a property, and only when the property's range is a composed sub-field of the **domain object**.

```
domain 1 object ------- scheme / property1 ------->  range 1 object
range 1 object ---------- rdf:type --------> range 1 class
range 1 object = domain 2 object --------- property2 ---------> range 2 value/object
range 1 object = domain 2 object --------- property3 ---------> range 2 value/object
```

parmentf:
And that is what is expressed in b3aed75:

```
<http://uri/1/compose/propa> rdf:type <http://class/2> .
<http://uri/1/compose/propa> <http://property/b> "value 2" .
<http://uri/1> <http://property/a> <http://uri/1/compose/propa> .
```

because `uri/1` has a property `property/a` which range class is `class/2` and which range object is `uri/1/compose/propa`, a composed subfield (composed of only property: `property/b`, which range value is `"value 2"`).

parmentf:
Plan, to take these rules into account:
- [ ] rewrite `mergeCompose.spec.js`
- [ ] rewrite `mergeCompose`
- [ ] review `merge`, to see if there remains no unwanted cases (classes on a non-composed field, etc.)

src/api/statements/JSONLDObject/mergeClasses.js
touv: Note that exist this statement : https://github.com/Inist-CNRS/lodex/blob/master/src/api/statements/LodexConfig.js

so we can not import config. We just need to test if input data contains $context. like thishttps://github.com/Inist-CNRS/lodex/blob/master/src/api/statements/LodexSetField.js#L14

PR #463, opened 2017-10-27
Write tests for exportExtendedNquads

parmentf:


parmentf:
The tests were integrated in #846

PR #464, opened 2017-10-27
[RFR] Drop min & raw format

touv:
it's not standard, so it's routine


parmentf:
You did not remove `src/api/exporters/exportMin.spec.js`, making the tests fail.

parmentf:
Wouldn't that break lodex-resource format?

Oh! No! I see you already moved it:
https://github.com/Inist-CNRS/lodex/blob/master/src/app/js/formats/lodex-resource/Component.js#L35

touv:
thanks for the review, i improve again this PR, to drop useless statements.
I have patched "resource" format, to use another route, because "min" export doesn't work as expected

PR #467, opened 2017-11-14
[RFR] Update deps

ThieryMichel:


fzaninotto (approved):

src/app/js/test.spec.js
```diff
@@ -1,2 +1,7 @@
+import Enzyme from 'enzyme';
```
fzaninotto: This should not be in a `.spec.js` file, which already matches mocha's test files pattern

fzaninotto (changes requested):

fzaninotto:
tests fail now...

PR #478, opened 2017-11-22
[RFM] enhancement: allow user to choose the parser at upload

touv:


src/api/controller/api/upload.js
```diff
@@ -18,12 +18,12 @@ import {
     createReadStream,
 } from '../../services/fsHelpers';
 
-export const getParser = (type) => {
-    if (!loaders[type]) {
-        throw new Error(`Unsupported document type: ${type}`);
+export const getParser = (parserName) => {
+    console.log('parserName', parserName);
```
fzaninotto: console.log spotted

fzaninotto (changes requested):

src/api/controller/api/upload.js
touv: I don't understand why the ./pre-commit script isn't executed... I think it should send me a warning for this type of mistake

fzaninotto:
needs rebase

PR #485, opened 2017-11-27
[RFR] Replace ISTEX ID URIs with ARK URIs

parmentf:
As there is an ARK for every ISTEX document, we can now replace local ISTEX IDs in documents' URIs with persistent ARKs.


parmentf:
Uh! Oh! I think I forgot to change the query, to include `arkIstex`.

See https://github.com/Inist-CNRS/lodex/blob/master/src/api/statements/scroll.js#L93-L95

parmentf:
Strange, the Travis CI build passes for `continuous-integration/travis-ci/push`, but no more for `continuous-integration/travis-ci/pr`.

It's been twice the same test that failed (and it is not _a priori_ concerned with this PR):

```
1) upload
       parseRequest
         should put info from resumable parsed request into ctx.resumable:
     TypeError: Cannot read property 'match' of undefined
      at exports.parseRequest (src/api/controller/api/upload.js:75:38)
      at <anonymous>
```

Issue #369, opened 2017-06-12
Export multi value fields to multiple triple


touv:
When a field value is an array we should transform each item in triple

touv:
![image](https://user-images.githubusercontent.com/15609/28019400-1f3bc718-6581-11e7-9373-86d1bb54557f.png)
![image](https://user-images.githubusercontent.com/15609/28019418-28004e96-6581-11e7-9dc6-652704170029.png)

parmentf:
If a field (of a resource, which uri is `http://uri`) is an array: `['a','b','c']` (with a scheme `http://scheme`), export as many triples as the number of items in the array:

```ttl
<http://uri> <http://scheme> "a" .
<http://uri> <http://scheme> "b" .
<http://uri> <http://scheme> "c" .
```

QuanticPotatoes:
Lodex can actually export an array.
In ***nquads*** : 
```xml
<http://localhost:3000/uid:/9XZPW6KN> <http://schema.org/actor> "Choi Min-sik" .
<http://localhost:3000/uid:/9XZPW6KN> <http://schema.org/actor> "Morgan Freeman" .
<http://localhost:3000/uid:/9XZPW6KN> <http://schema.org/actor> "Scarlett Johansson" .
```
and ***turtle*** : 
```ttl
schema:actor "Choi Min-sik", "Morgan Freeman", "Scarlett Johansson";
```

parmentf:
I just checked. It works.

PR #487, opened 2017-11-29
[RFR] Add TSV-double-quotes loader

parmentf:
This new loader uses " as a quote.

The name of the PR refers to #369, but I came to check, and this does already work.
Thanks @QuanticPotatoes ! ;)


parmentf:
Let's write a test for that.

fzaninotto (approved):
Yes, a test looks necessary. Perhaps some jsDoc in the loader script as well?

fzaninotto (approved):

PR #504, opened 2017-12-06
[RFR] change graph summary to be a drop down menu

ThieryMichel:


src/app/js/public/graph/GraphSummary.js
```diff
@@ -1,43 +1,69 @@
...
+        iconButtonElement={
+            <FlatButton primary>
+                {polyglot.t('graph_list').toUpperCase()}
+            </FlatButton>
+        }
+    >
+        <MenuItem
+            value=""
+            primaryText={
+                <Link style={styles.link} to="/graph">
```
fzaninotto, line 39: I think the practice is to use `<MenuItem component={Link} to="/graph" primaryText="dataset" />` instead

fzaninotto (changes requested):

src/app/js/public/graph/GraphSummary.js
ThieryMichel: nope, not possible, maybe in material-ui 1.0, but I am in material-ui 0.19.

PR #509, opened 2017-12-08
[RFR] New transformers

touv:
- trim
- capitalize
- uniq

### DIFF
--- a/src/common/transformers/CAPITALIZE.js
+++ b/src/common/transformers/CAPITALIZE.js
@@ -1,22 +1,16 @@
-export const capitalizeString = value => {
-    if (!value) {
-        return null;
-    }
+import smartMap from './smartMap';
 
-    if (Array.isArray(value)) {
-        return value.map(capitalizeString);
-    }
+export const capitalizeString = value =>
+    typeof value === 'string'
+        ? value.charAt(0).toUpperCase() + value.slice(1).toLowerCase()
+        : value;
 
-    if (typeof value === 'object') {
-        return null;
-    }
-    return value.charAt(0).toUpperCase() + value.slice(1).toLowerCase();
-};
+export const capitalize = smartMap(capitalizeString);
 
 const transformation = () => value =>
     new Promise((resolve, reject) => {
         try {
-            resolve(capitalizeString(value));
+            resolve(capitalize(value));
         } catch (error) {
             reject(error);
         }
--- a/src/common/transformers/CAPITALIZE.js
+++ b/src/common/transformers/CAPITALIZE.js
@@ -1,0 +1,31 @@
+export const capitalizeString = value => {
+    if (!value) {
+        return null;
+    }
+
+    if (Array.isArray(value)) {
+        return value.map(capitalizeString);
+    }
+
+    if (typeof value === 'object') {
+        return null;
+    }
+    return value.charAt(0).toUpperCase() + value.slice(1).toLowerCase();
+};
+
+const transformation = () => value =>
+    new Promise((resolve, reject) => {
+        try {
+            resolve(capitalizeString(value));
+        } catch (error) {
+            reject(error);
+        }
+    });
+
+transformation.getMetas = () => ({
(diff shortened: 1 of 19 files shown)


src/common/transformers/TRIM.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+
+    it('should return trim value of each item in array', () => {
+        expect(trimString([' hello   ', '\nworld\r\n'])).toEqual([
+            'hello',
+            'world',
+        ]);
+    });
+
+    it('should return null if value is literal', () => {
```
parmentf: if value is NOT a literal?

src/common/transformers/UNIQ.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+
+    it('should return uniq value of each item in array', () => {
+        expect(uniqArray(['hello', 'world', 'hello', 'world'])).toEqual([
+            'hello',
+            'world',
+        ]);
+    });
+
+    it('should return null if value is literal', () => {
```
parmentf: if value is NOT a literal, or better: if value is an object.

touv: I didn't change text from uppercase.spec.js, it's a copy/paste

src/common/transformers/TRIM.spec.js
ThieryMichel: no, the text is right, { a: 'hello ', b: 'world ' } is a literal. But maybe `is not a string` would be even better.

src/common/transformers/UNIQ.spec.js
ThieryMichel: The text is right here too. And is `not an array` is even better

src/common/transformers/UNIQ.js
```diff
@@ -0,0 +1,32 @@
+export const uniqArray = value => {
+    if (!value) {
+        return null;
+    }
+
+    if (Array.isArray(value)) {
+        return value.filter((v, i, a) => a.indexOf(v) === i);
```
ThieryMichel: You should use lodash.uniq instead, it do the same and is bullet proof.

src/common/transformers/CAPITALIZE.js
```diff
@@ -0,0 +1,31 @@
+export const capitalizeString = value => {
+    if (!value) {
+        return null;
+    }
+
+    if (Array.isArray(value)) {
+        return value.map(capitalizeString);
+    }
+
+    if (typeof value === 'object') {
```
ThieryMichel: A suggstion: wouldn't it better if you capitalized the value of the literal ?
```js
Object.keys(value).reduce((acc, key) => ({
    ...acc,
    [key]: capitalizeString(value[key]),
}));
```

src/common/transformers/TRIM.js
```diff
@@ -0,0 +1,32 @@
+export const trimString = value => {
+    if (!value) {
+        return null;
+    }
+
+    if (Array.isArray(value)) {
+        return value.map(trimString);
+    }
+
+    if (typeof value === 'object') {
```
ThieryMichel: Suggestion: you could trim value of literal (see capitalize)

src/common/transformers/CAPITALIZE.js
```diff
@@ -0,0 +1,31 @@
+export const capitalizeString = value => {
+    if (!value) {
```
ThieryMichel: Be it capitalize, uniq, or trim, the only thing that change is the function applied to the value.
Please refactor in order to reuse the common logic.

eg:

```js
const transform = fn => value => {
    if(!value) {
        return null;
    }
     ...
    return fn(value);
}
```

ThieryMichel: This way you only have to test transform, and fn, and this will avoid all test copy paste.

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #518, opened 2017-12-13
[RFR] misc facet changes

ThieryMichel:


src/app/js/public/facet/AppliedFacetList.js
```diff
@@ -26,18 +30,29 @@ export const AppliedFacetListComponent = ({ facets }) => (
...
+    clearAll: PropTypes.func.isRequired,
 };
 
 const mapStateToProps = state => ({
     facets: fromFacet.getAppliedFacetList(state),
 });
 
-export default connect(mapStateToProps)(AppliedFacetListComponent);
+const mapDispatchToProps = {
+    clearAll: () => clearFacet(),
```
fzaninotto, line 53: why use a different name?

src/app/js/public/graph/sagas.js
```diff
@@ -76,15 +83,22 @@ export function* handleLoadChartDataRequest({ payload: { field } }) {
...
+    yield put(loadChartDataSuccess({ name, data: null }));
 }
 
 export default function*() {
     yield fork(function*() {
-        yield takeEvery(
-            [LOAD_CHART_DATA, TOGGLE_FACET_VALUE, CLEAR_FACET, APPLY_FILTER],
+        yield throttle(
+            500,
+            [
```
fzaninotto: I can't find in the `redux-saga` doc the part about `throttle` taking an array as second argument. Can you add a link in a JS comment?

fzaninotto (changes requested):

src/app/js/public/facet/AppliedFacetList.js
ThieryMichel: to emphasize that this call will clear all facet and not just one

ThieryMichel: And also to not pass click event in the call.

fzaninotto (approved):

PR #527, opened 2017-12-19
[RFR] Fix routine order by reload chartData on field change and allow to edit field from graph page

ThieryMichel:


src/api/statements/LodexParseQuery.spec.js
```diff
@@ -0,0 +1,67 @@
+import expect from 'expect';
+
+import LodexParseQuery from './LodexParseQuery';
+
+describe.only('LodexParseQuery', () => {
```
fzaninotto: `.only` spotted

fzaninotto (changes requested):

fzaninotto (approved):

PR #536, opened 2018-01-02
[RFR] Heatmap

ThieryMichel:
capture du 2018-01-02 17-29-59


src/app/js/formats/heatmap/HeatMapView.js
```diff
@@ -0,0 +1,224 @@
...
+                                            className={css(styles.rotateSpan)}
+                                        >
+                                            {xKey}
+                                        </span>
+                                    </div>
+                                </td>
+                                {yAxis.map(yKey => (
+                                    <td
+                                        className={css(styles.td)}
+                                        key={yKey}
```
fzaninotto: not good, you'll end up with duplicate keys as it's repeated for every line. I advise you to use a compound of `xKey` and `yKey` here

src/app/js/formats/heatmap/parseChartData.spec.js
```diff
@@ -0,0 +1,101 @@
+import expect from 'expect';
+
+import { mapSourceToX, mapTargetToX } from './parseChartData';
+
+describe.only('parseChartData', () => {
```
fzaninotto: `.only` spotted!

fzaninotto (changes requested):

fzaninotto:
Seing the screenshot, I worry about the horizontal scroll bar (which is never a good ui). Can't the columns width adjust to the available width?

ThieryMichel:
If we make the columns smaller we won't have enough place for the label, also we can flip the two axes, it's in the option of the format. Granted it won't fix all case.

capture du 2018-01-03 18-24-12

src/app/js/formats/heatmap/HeatMapView.js
ThieryMichel: I thought that as long as all the key in the array of component are unique, it's good, even if there is an element with the same key elsewhere. Also react print no duplicate key error.
I guess it cannot hurt, just not sure it is really needed.

fzaninotto:
I think it makes more sense with the years as columns. what do the PO say?

fzaninotto (approved):

PR #543, opened 2018-01-04
[RFR] clear facet when changing page

ThieryMichel:


src/app/js/public/facet/sagas.js
```diff
@@ -28,9 +30,18 @@ export function* handleLoadFacetValuesRequest({ payload: { name } }) {
     return yield put(loadFacetValuesSuccess({ name, values }));
 }
 
+export function* clearFacetSaga({ payload: { action } }) {
+    if (action === 'POP') {
```
fzaninotto, line 34: please use a constant instead of a string here

src/app/js/public/graph/sagas.js
```diff
@@ -83,7 +83,6 @@ export default function*() {
     yield throttle(
         500,
         [
-            LOAD_CHART_DATA,
```
fzaninotto, line 86: rogue change?

fzaninotto (changes requested):

src/app/js/public/graph/sagas.js
ThieryMichel: This action was triggering the saga twice (here in throttle and later in a takeEvery)

src/app/js/public/facet/sagas.js
ThieryMichel: No constant exist for this value (it come from react-router-redux)

fzaninotto (approved):

PR #546, opened 2018-01-05
[RFR] Fix field validation

ThieryMichel:


fzaninotto (changes requested):
You should add a unit test for the case you just fixed

fzaninotto (approved):

PR #547, opened 2018-01-05
[RFR] Table ontology

ThieryMichel:


src/app/js/fields/ontology/Ontology.js
```diff
@@ -92,22 +33,82 @@ export class OntologyComponent extends Component {
         handleChangePosition({ newPosition: newIndex, oldPosition: oldIndex });
     };
 
+    constructor(props) {
+        super(props);
+        this.state = { filter: 'all' };
+    }
+
+    handleFilterChange = (_, __, filter) =>
+        this.setState(state => ({
```
fzaninotto: yu can use the simpler `this.setState({ filter })` in that case

src/app/js/fields/ontology/OntologyFieldList.js
```diff
@@ -0,0 +1,50 @@
...
+import { TableBody, TableRow } from 'material-ui/Table';
+
+import OntologyField from './OntologyField';
+import { COVER_DATASET } from '../../../../common/cover';
+
+const OntologyFieldList = SortableContainer(({ items, filter }) => (
+    <TableBody displayRowCheckbox={false}>
+        {items
+            .filter(
+                field =>
```
fzaninotto: extract the filter function, so that it's unit testeable, and appears as a simple name here (better for readability). same for the other filter function below.

fzaninotto (changes requested):

fzaninotto (approved):

PR #550, opened 2018-01-08
[RFR] Upload dataset update

ThieryMichel:


fzaninotto (approved):

src/api/controller/api/upload.js
```diff
@@ -138,37 +210,18 @@ export const checkChunkMiddleware = async ctx => {
...
-    };
-};
-
 const app = new Koa();
 
 app.use(prepareUpload);
 
 app.use(koaBodyParser());
 app.use(route.post('/url', uploadUrl));
+app.use(route.post('/url', saveParsedStream));
```
fzaninotto: Would better be used as a function call at the end of the `uploadUrl` route rather than as another middleware

src/api/services/getAddUriTransformer.js
```diff
@@ -0,0 +1,14 @@
+import omit from 'lodash.omit';
+
+import getDocumentTransformer from './getDocumentTransformer';
+
+export const addUriToDoc = getUriDoc => async doc => ({
```
fzaninotto: this one is a bit cryptic, could you add some jsDoc to explain its use?

fzaninotto (changes requested):
great job! Lots of tests, good refactoring.

fzaninotto (approved):

PR #569, opened 2018-01-18
[RFR] Scroll

ThieryMichel:


src/app/js/Root.js
```diff
@@ -7,7 +7,10 @@ import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
 const Root = ({ store, routes, history }) => (
     <Provider {...{ store }}>
         <MuiThemeProvider>
-            <Router {...{ history, routes }} />
+            <Router
+                {...{ history, routes }}
+                onUpdate={() => window.scrollTo(0, 0)}
```
fzaninotto: please extract since it's not dynamic - this will avoid re-renders due to a failed props equality check.

src/app/js/public/facet/sagas.js
```diff
@@ -37,11 +38,20 @@ export function* clearFacetSaga({ payload: { action } }) {
     yield put(clearFacet());
 }
 
+const scrollToTop = () =>
+    window.scroll({ top: 0, left: 0, behavior: 'smooth' });
```
fzaninotto: is it the same used in src/app/js/Root.js?

Also, please check the existence of `window` before calling `scroll` - this will ease unit tests in Node.js

fzaninotto (changes requested):

src/app/js/public/facet/sagas.js
ThieryMichel: It is not the same since here I need a smooth animation, whereas on route change I need it to be instant. But it still could be refactored into a single helper function :)

fzaninotto (approved):

PR #574, opened 2018-01-19
[RFR] Semantic post publish

ThieryMichel:


src/app/js/fields/FieldCompletes.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
```
fzaninotto: the naming is weird... I'd prefer `FieldComplement` or `FieldAnnotation`. If it's too cumbersome, keep it like that, but try to avoid using verbs in active form in component names.

src/app/js/fields/FieldCompletes.js
```diff
@@ -0,0 +1,59 @@
...
+    complete: {
+        paddingLeft: '40px',
+    },
+    header: {
+        lineHeight: '48px',
+        paddingLeft: 0,
+    },
+};
+
+const FieldComplete = ({ fields, p: polyglot }) => {
```
fzaninotto: not the same as the field name...

src/app/js/fields/FieldCompletes.js
```diff
@@ -0,0 +1,59 @@
...
+    const otherFieldsMenuItems = fields.map(f => (
```
fzaninotto: what's the interest of extracting it from the JSX if you don't reuse it or test it?

src/app/js/fields/FieldComposedOf.js
```diff
@@ -0,0 +1,188 @@
...
+                                index
+                            }`}
+                            onChange={handleSelectColumn(index)}
+                            style={styles.select}
+                            hintText={polyglot.t('select_a_column')}
+                            value={col}
+                            required
+                        >
+                            {fields.map(f => (
+                                <MenuItem
```
fzaninotto: The JSX tends to be too indented and too long, split it into smaller components.

src/app/js/fields/FieldComposedOf.js
```diff
@@ -0,0 +1,188 @@
...
+
+const mapDispatchToProps = (dispatch, { FORM_NAME = FIELD_FORM_NAME }) => ({
+    onChange: composedOf => {
+        dispatch(change(FORM_NAME, 'composedOf', composedOf));
+    },
+});
+
+export default compose(
+    connect(mapStateToProps, mapDispatchToProps),
+    withHandlers({
```
fzaninotto: not a big fan, this component is complex enough to justify using the class syntax, and therefore avoid the usage of `withHandlers`

fzaninotto (changes requested):

fzaninotto (approved):
I just don't understand why you need a saga to call `validateAddedField`, as it has no side effect and it's synchronous.

PR #580, opened 2018-01-23
[RFR] Misc fix

ThieryMichel:


src/app/js/formats/distributionChart/global-bar-chart/BarChartView.js
```diff
@@ -156,13 +149,19 @@ const mapStateToProps = (state, { field, chartData }) => {
         categoryMargin: parseInt(categoryMargin),
     });
 
+    const height =
```
fzaninotto, line 152: I don't understand the interest of putting that calculation in `mapStateToProps` instead of `render`.

fzaninotto (approved):

PR #581, opened 2018-01-23
[RFR] Authentication

ThieryMichel:


fzaninotto (commented):
I don't know how the feature was described, but you should always compare password hashes and not passwords.

fzaninotto (approved):

PR #621, opened 2018-02-04
[RFR] to avoid to lock /login if userAuth is defined

touv:
to test in different cases before merge

### DIFF
--- a/src/api/controller/front.js
+++ b/src/api/controller/front.js
@@ -200,7 +200,7 @@ if (config.userAuth) {
     );
 
     app.use(async (ctx, next) => {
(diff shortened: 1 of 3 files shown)


src/api/controller/front.js
```diff
@@ -203,7 +203,8 @@ if (config.userAuth) {
         if (
             !ctx.state.cookie &&
             ctx.request.url !== '/login' &&
-            ctx.request.url !== '/index.js'
+            ctx.request.url !== '/index.js' &&
+            ctx.request.url !== '/index.js.map'
```
parmentf: Oh! I did not understand! Thanks.

PR #622, opened 2018-02-04
[RFR] Try avoid max size warning

touv:
with the existing model, the warning persists...


parmentf:
To explain a bit, this warning was associated with a wrong behaviour of the `More` button in ResourcesGrid.

PR #626, opened 2018-02-08
[RFR] a specific title & description for each pages

touv:
Now the title and description are chosen with the Syndication menu. So …
any field could be used as a title or description. In addition, we can
choose dataset title & description as the same way.


src/app/js/fields/selectors.js
```diff
@@ -121,16 +120,19 @@ const getAllListFields = createSelector(getCollectionFields, fields =>
     fields.filter(f => !f.composedOf),
 );
 
-const getOverViewCol = type => fields => {
-    const result = fields.filter(f => f.overview === type);
+const findFieldWithOverviewID = id => fields => {
+    const result = fields.filter(f => f.overview === id);
```
parmentf, line 124: Why use `filter`, while there is also `find`.
http://devdocs.io/javascript/global_objects/array/find

src/app/js/i18n/translations/fr.js
```diff
@@ -167,6 +167,11 @@ export default {
     show_publication_errors: 'Erreurs',
     model: 'Modèle',
     overview: 'Syndication',
+    overviewNone: 'Aucun',
+    overviewTitle: 'RESSOURCE - Titre',
+    overviewDescription: 'RESSOURCE - Description',
+    overviewDatasetTitle: 'COLLECTION - Titre',
```
parmentf: What?!? But there is also a `cover_collection` key which translation is `Différent pour chaque ressource`. Collection is a  cover level, doesn't it?

In fact, `error_cover_invalid_cover` says `Cover must be either dataset or collection`. That means that Collection is different from Dataset.

One can't translate `Dataset` with `Collection`.

parmentf:
Currently, the /graph page does not have a title, and thus the page's title in the browser is its URL.

parmentf:
_RESOURCE Title_ and _RESOURCE Description_ are visible, through a badge in the ontology (view model):
image

it would be fine if the _DATASET_ side had the same badges.

src/app/js/fields/selectors.js
touv: just now, I try to use find and I don't no why but it's can not works : 

```
const findFieldWithOverviewID = id => fields =>
    fields.find(f => f.overview === id);
```

so I prefer to leave as it is

src/app/js/i18n/translations/fr.js
touv: perhaps, the best way, is to do not translate DATASET & RESOURCE prefix in french...

src/app/js/fields/selectors.js
parmentf: But that's a `name`, that should be returned, not a field. (See line 125 below).

PR #629, opened 2018-02-13
[RFR] Do not use config in the webpack bundle

touv:
- there is no longer to rebuild at startup
- there is no password stored in the webpack bundle !!!!!
- there allows to start lodex on small server


parmentf:
To change `schemeForDatasetLink`, we could use a dataset-level, not displayed by default, which name would be `schemeForDatasetLink`...

src/app/js/formats/istex/ListComponent.js
```diff
@@ -5,7 +5,7 @@ import memoize from 'lodash.memoize';
 import { REJECTED } from '../../../../common/propositionStatus';
 import { field as fieldPropTypes } from '../../propTypes';
 
-const istexApiUrl = process.env.ISTEX_API_URL;
+const istexApiUrl = 'https://api.istex.fr/document';
```
parmentf, line 8: Too bad you have to declare it three times.

config/default.js
```diff
@@ -34,7 +10,5 @@ module.exports = {
         expiresIn: 10 * 3600, // 10 hours
     },
     hostname: 'http://data.istex.fr/',
```
parmentf: Shouldn't it be https://data.istex.fr (https, no more http)?

PR #639, opened 2018-03-06
[RFR] change default filter in model view (backoffice)

touv:


parmentf (changes requested):
Good idea.

But, IMHO, there is a lack of label to the dropdown list, explaining that fields are filtered.

parmentf (approved):

parmentf:
I would have put a supplementary label, in front of the list, but that's enough.

**Note**: the labels are too long to display what kined of fields are displayed, but the title below says it, anyway.

image

PR #643, opened 2018-03-07
[RFR] Add an empty div to LodexResource

parmentf:
In order to be able to add "To know more" link.


parmentf:
[Ajouter une div vide dans LodexResource](https://trello.com/c/DCdebUhN/105-ajouter-une-div-vide-dans-lodexresource)

PR #644, opened 2018-03-07
[RFR] try to improve table (short column name and small uri column)

touv:


parmentf (commented):
It's still a bit weird, when the column is much larger than the truncated label of the column.
image

PR #646, opened 2018-03-12
[RFR] use the standard KOA error handler

touv:


touv:
see https://trello.com/c/AY8CxJyK

touv:
I suspect the error handler to  send data before the headers

parmentf:
Hard to test. I trust you. ;P

PR #656, opened 2018-03-23
[RFR] introduce 4 new transformers

touv:
- FORMAT
 - REPLACE
 - REMOVE
 - LOWERCASE


src/common/transformers/PREFIX.js
```diff
@@ -1,34 +1,10 @@
...
-    if (!value) {
-        return str;
-    }
+import { transformerWithArg } from './transformer';
 
-    if (typeof value === 'string') {
-        return str.concat(value);
-    }
+export const prefix = (value, str) =>
+    typeof value === 'string' ? str.concat(value) : value;
```
parmentf, line 4: This prevents to use PREFIX on an array, we have to update [the documentation](https://www.gitbook.com/book/lodex/lodex-user-documentation/edit#/edit/master/Administration/Mod%C3%A8le/Transformers/PREFIX.md).

src/common/transformers/SUFFIX.js
```diff
@@ -1,34 +1,10 @@
...
-    if (!value) {
-        return str;
-    }
+import { transformerWithArg } from './transformer';
 
-    if (typeof value === 'string') {
-        return value.concat(str);
-    }
+export const suffix = (value, str) =>
+    typeof value === 'string' ? value.concat(str) : value;
```
parmentf, line 4: Same problem as with SUFFIX: no more applicable on an array.
We have to update [the documentation](https://www.gitbook.com/book/lodex/lodex-user-documentation/edit#/edit/master/Administration/Modèle/Transformers/SUFFIX.md).

src/common/transformers/UPPERCASE.js
```diff
@@ -1,27 +1,9 @@
-import smartMap from './smartMap';
+import { transformer } from './transformer';
 
-export const valueToUpperCase = value => {
-    if (!value) {
-        return null;
-    }
+export const valueToUpperCase = value =>
+    typeof value === 'string' ? value.toUpperCase() : value;
```
parmentf, line 4: Here too, we have to update [the documentation](https://www.gitbook.com/book/lodex/lodex-user-documentation/edit#/edit/master/Administration/Modèle/Transformers/UPPERCASE.md).

parmentf (changes requested):
Nice refactoring for `transformer` instead of `smartMap`.

Write tests for LOWERCASE, please.

And make sure that we will update documentation.

src/common/transformers/REPLACE.js
```diff
@@ -0,0 +1,24 @@
+import { transformerWithTwoArgs } from './transformer';
+
+export const replace = (value, from, by) =>
+    typeof value === 'string' ? value.split(from).join(by) : value;
```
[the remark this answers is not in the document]
touv: http://jsben.ch/LFfWA

src/common/transformers/LOWERCASE.js
```diff
@@ -0,0 +1,14 @@
...
+
+const transformation = () => value => transformer(valueToLowerCase, value);
+
+transformation.getMetas = () => ({
+    name: 'LOWERCASE',
+    type: 'transform',
+    args: [],
+});
+
+export default transformation;
```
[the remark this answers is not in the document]
touv: is it very useful ?

src/common/transformers/PREFIX.js
touv: We can use it on an array, like all others transformers :
BOOLEAN, CAPITALIZE, LOWERCASE, NUMBER, STRING, TRIM, UPPERCASE, PREFIX, SUFFIX, REPLACE

See 
https://github.com/Inist-CNRS/lodex/blob/a6b2bce09288a7ecf49b4d7ae2bb9d433adbd714/src/common/transformers/transformer.js#L4-L8

touv: but the documentation will sould be update, because, now, all "string" transformers will be apply on each item of the array. It's more constitent et more usefull.
If we need to add item on array, i think it's better to add a specific transfomer

src/common/transformers/LOWERCASE.js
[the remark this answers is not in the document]
parmentf: Maybe not *very* useful in this case, but... if every `.js` file could have a `.spec.js` counterpart, it would be easy to know that everything is tested.

src/common/transformers/PREFIX.js
parmentf: Oh! I did not notice that.
Smarter than I thought. :)

parmentf (approved):

src/common/transformers/LOWERCASE.js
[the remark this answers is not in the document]
touv: I added it ;-)

PR #658, opened 2018-03-28
[RFR] Export model ready

parmentf:
Add an export of the model, which can be directly used with exported data.

(the first time, you have to adapt the model, because all transformers are useless, only use the one which pick the value from a column)

I call it "model ready" because the model is ready to be applied to the exported data. Maybe I should call it "ready model".


parmentf:
[ajouter l'export du modèle adapté aux données exportées](https://trello.com/c/NFcv0MVY/96-ajouter-lexport-du-mod%C3%A8le-adapt%C3%A9-aux-donn%C3%A9es-export%C3%A9es)

touv (approved):

PR #661, opened 2018-04-17
[RFR] modify formats

chaBiselx:


parmentf:
OK. Since we have tested it manually, we know my proposition works.

As soon as it possible, don't disable eslint, it often is a warning about a potential problem.

PR #672, opened 2018-05-29
[RFR] Affichage d'image si possible (SPARQL - TEXT)

chaBiselx:


src/app/js/formats/sparql/SparqlTextField/SparqlView.js
```diff
@@ -53,12 +54,24 @@ const styles = {
         color: 'rgb(158, 158, 158)',
         fontSize: '1.3rem',
     },
+    imgDefault: {
+        height: '200px',
```
parmentf: If you set the height *and* the width of the image, it will always be square!
I'd prefer to set only the width, so that it won't be too large (the height is less important, but in that way, the ratio width/height will be good).

parmentf: Even better: set `max-width`, and not `width` (maybe larger then 200px, I'd say maximum 900px).

parmentf (changes requested):

parmentf (approved):

PR #677, opened 2018-06-07
[RFR] SPARQL - TEXTE sous format liste

chaBiselx:


src/app/js/formats/sparql/SparqlTextField/SparqlView.js
```diff
@@ -53,9 +54,42 @@ const styles = {
...
+        textDecoration: 'none',
+        margin: 0,
+    },
 };
 
 export class SparqlTextField extends Component {
+    ifArray = result => {
+        const { className, sparql } = this.props;
+        const temp = clonedeep(result);
+        if (temp[1].value.includes(sparql.separator)) {
```
parmentf, line 73: Aren't there any cases where `temp[1]` is undefined?
And thus where `temp[1].value` throws an error (I didn't succeed in triggering such a case, but I don't know the structure of `obj` from `topairs`).

chaBiselx: toPairs transform a json objet in array : 
attribute => [0] (the label)
value => [1] (a json object) 
image

PR #684, opened 2018-06-21
[RFR] Améliorer SPLIT pour eviter les chaînes vides

chaBiselx:


src/common/transformers/SPLIT.js
```diff
@@ -8,8 +8,8 @@ export const split = (value, separator) => {
     if (typeof value !== 'string') {
         throw new Error('Invalid value: need a string');
     }
-
-    return value.split(separator).map(x => x.trim());
+    let removeEmpty = value.split(separator).map(x => x.trim());
```
parmentf: A variable name, again ;)

I would have called this variable `splitted`. `removeEmpty` suggest an action, and thus a function.

parmentf (changes requested):

parmentf (approved):

PR #691, opened 2018-09-18
[RFR] Show empty resource fields for admin only

Kmaschta:
[Trello card](https://trello.com/c/4Su43hIy)


src/app/js/public/resource/Detail.js
```diff
@@ -114,13 +113,28 @@ const styles = {
...
 };
 
-export const DetailComponent = ({ fields, resource, title, description }) => {
+export const shouldDisplayField = (resource, isAdmin) => field => {
+    if (isAdmin) {
+        return true;
+    }
+
+    const isEmptyValue = !resource[field.name];
+    return !isEmptyValue || Boolean(field.composedOf);
```
ThieryMichel, line 122: composedOf, is of the form { separator: 'a delimiter', fields: ['fieldName',...] }, it show field that regroup 2 or more other fields.
So, what if all the composed field are empty ?
Also it can be recursive.

ThieryMichel (dismissed):

PR #699, opened 2018-09-24
[RFR] Fix slow mongo

ThieryMichel:


src/api/controller/api/field.js
```diff
@@ -131,6 +124,16 @@ export const importFields = getUploadedFieldsImpl => async ctx => {
     ctx.status = 200;
 };
 
+export const reorderField = async ctx => {
+    const { fields } = ctx.request.query;
+    ctx.body = await Promise.all(
+        Object.values(fields).map((name, index) =>
+            ctx.field.findOneAndUpdate({ name }, { $set: { position: index } }),
```
Kmaschta: I would ensure some validations here IMO
Like, at least, check that the URI field is always at position 0
(if its needed)

src/app/js/user/index.js
```diff
@@ -334,6 +335,14 @@ export const getExportPublishedDatasetRequest = (
         url: `/api/export/${type}?${queryString}`,
     });
 
+export const getReorderFieldRequest = (state, fields) =>
+    getRequest(state, {
+        method: 'PUT',
+        url: `/api/field/reorder?${qs.stringify({
```
Kmaschta: Using query string in this case seems so weird to me..
Any particular reason not to use JSON body?

Kmaschta (changes requested):

src/api/controller/api/field.js
```diff
@@ -131,6 +129,49 @@ export const importFields = getUploadedFieldsImpl => async ctx => {
     ctx.status = 200;
 };
 
+export const reorderField = async ctx => {
+    const { fields } = ctx.request.body;
+
+    const fieldsData = await Promise.all(fields.map(ctx.field.findOneByName));
```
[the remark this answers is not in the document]
Kmaschta: What's the issue with the sort?

Kmaschta:
Rebase needed

Kmaschta (approved):

Kmaschta (approved):

PR #700, opened 2018-09-25
[RFR] Remove LINK transformer

Kmaschta:


ThieryMichel (approved):
Seems good to me.
Now we can simplify the publishing in a next PR

PR #712, opened 2018-10-01
[RFR] Preparations for new graphics

touv:


Kmaschta (approved):

src/common/transformers/GET.js
```diff
@@ -0,0 +1,20 @@
+import _get from 'lodash.get';
+import { transformerWithArg } from './transformer';
+
+export const get = (value, path) => {
+    if (!value) {
```
ThieryMichel: I think it's not needed as _.get accept a default return value as the third parameter, and it should cover the case when we have no value.

src/common/transformers/GET.spec.js
```diff
@@ -0,0 +1,9 @@
+import expect from 'expect';
+
+import { get } from './GET';
+
+describe('GET', () => {
+    it("should return key's value from object", () => {
+        expect(get({ a: { aa: 1 }, b: 2 }, 'a.aa')).toEqual(1);
+    });
```
ThieryMichel, line 8: test with no value, incorrect path, and no path too.

ThieryMichel (changes requested):

src/common/transformers/GET.js
touv: ok, this test is useless, i will drop it

src/common/transformers/GET.spec.js
touv: I do not think it is useful to thoroughly test the lodash function, the value is elsewhere

Kmaschta (approved):

PR #717, opened 2018-10-02
[RFR] do not set progress when publishing facet outside of publication

ThieryMichel:


src/api/controller/api/field.js
```diff
@@ -58,7 +58,7 @@ export const putField = async (ctx, id) => {
         return;
     }
 
-    await ctx.publishFacets(ctx, [ctx.body]);
+    await ctx.publishFacets(ctx, [ctx.body], false);
```
Kmaschta, line 61: This `false` is no longer useful then

src/api/controller/api/publishFacets.spec.js
```diff
@@ -41,7 +41,7 @@ describe('publishFacets', () => {
     ];
 
     before(async () => {
-        await publishFacets(ctx, facetFields);
+        await publishFacets(ctx, facetFields, false);
```
Kmaschta, line 44: No need of false

Kmaschta (changes requested):

Kmaschta (approved):

PR #722, opened 2018-10-04
[RFR] Fix multiple value in facets

Kmaschta:
[Trello card](https://trello.com/c/NNqaVD1U)


cypress/integration/publication.spec.js
```diff
@@ -88,4 +88,27 @@ describe('Dataset Publication', () => {
             cy.contains('Row 2').should('be.visible');
         });
     });
+
+    describe('Facets', () => {
+        it('should allow to have facets with multiples values in them', () => {
```
ThieryMichel, line 93: how about testing when facet has simple values too ?

ThieryMichel (changes requested):

cypress/integration/publication.spec.js
Kmaschta: I'll do ! On another PR

ThieryMichel (approved):

PR #723, opened 2018-10-04
[RFR] Fix *recursively* quick field edition

Kmaschta:
[Trello card](https://trello.com/c/W36trY4p)

## TODO
- [x] Make the fix
- [x] Ensure the same apply to the admin edition

### DIFF
--- a/src/api/controller/api/characteristic.js
+++ b/src/api/controller/api/characteristic.js
@@ -18,12 +18,46 @@ const isValueField = field =>
     get(field, 'transformers[0].operation') === 'VALUE';
 
 const updateFieldValue = async (ctx, field, value) => {
+    if (get(field, 'transformers[0].args[0].value') === value) {
+        return field;
+    }
+
     // lodash.set mutate the object, so we need to copy the object first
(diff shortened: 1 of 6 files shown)


src/api/controller/api/characteristic.js
```diff
@@ -18,14 +18,48 @@ const isValueField = field =>
...
+    return ctx.field.updateOneById(
         field._id,
         set(copiedField, 'transformers[0].args[0].value', value),
     );
 };
 
+const updateField = (ctx, requestedNewCharacteristics) => async field => {
+    const body = {};
+
+    if (isValueField(field)) {
```
ThieryMichel, line 36: how about adding a getFieldType function and then using switch ?

ThieryMichel (changes requested):

src/api/controller/api/characteristic.js
Kmaschta: I can't. What about the field is both VALUE and composedOf?

ThieryMichel (approved):

PR #727, opened 2018-10-05
[RFR] Preparations for new graphics3

touv:


src/api/controller/api/run.js
```diff
@@ -48,6 +48,15 @@ export const runRoutine = async (ctx, routineCalled) => {
         host: ctx.host,
         hostname: ctx.hostname,
     };
+    const fields = ctx.path
```
Kmaschta: Manually parse the request URL isn't recommended.
It's better to pass all the params from the router IMHO

```js
export const runRoutine = async (ctx, routineCalled, field1, field2) => {
    // ...
    const fields = [routineCalled, field1, field].filter(x => x);
}
```

touv: Yes, you're right. 
However, some older routines use ctx.path to detect the concerned fields.
I'm afraid to break something, I propose to change nothing to ensure backward compatibility

touv: oh... I understood, it's a new variable so ctx.path is still accessible... i'll patch

src/api/statements/LodexRunQuery.js
```diff
@@ -13,7 +13,7 @@ export const createFunction = mongoClientImpl =>
         }
         const from = this.getParam('from', data.$from || 'dataset');
         const query = this.getParam('query', data.query || {});
-        const limit = this.getParam('limit', data.limit || 10);
+        const limit = this.getParam('limit', data.limit || 1000000);
```
Kmaschta, line 16: Is this the real default value or just to debug?
If you confirm it's the correct value, I'll merge! :)

Kmaschta (approved):

src/api/statements/LodexRunQuery.js
touv: yes, it's the correct value

PR #736, opened 2018-10-12
 Enable maxSize, minValue, maxValue, orderBy for all charts 

touv:
(and enable new advanced routines.)

It's a big copy/paste  from barchart admin component, of course it's not lean, but I can't do better


src/app/js/formats/bubbleChart/BubbleAdmin.js
```diff
@@ -42,6 +50,30 @@ class BubbleAdmin extends Component {
         args: defaultArgs,
     };
 
+    setMaxSize = (_, maxSize) => {
+        const { params, ...args } = this.props.args;
```
ThieryMichel: refactor to add an helper function, and avoid to repeat the logic each time. Something like:
```js
const updateAdminParams = (key, value, props) => {
     const { params, ...args } = props.args;
     const newArgs = { ...args, params: { ...params, [key]: value } };
     props.onChange(newArgs);
}
```
And use it in all Admin component.

ThieryMichel (changes requested):

ThieryMichel (approved):

Kmaschta (approved):

PR #737, opened 2018-10-12
Why tests fails

touv:
some try ...


Kmaschta (commented):
Why not just update `statement` instead of creating a `statement2` ?

touv:
yes, the statement instruction has been updated in the new version of the package

PR #743, opened 2018-10-16
fix bad value for format fetching data

ThieryMichel:
do not try to fetch formatData if value is not a string, also use InvalidFormat in injectData

### DIFF
--- a/src/app/js/formats/injectData.js
+++ b/src/app/js/formats/injectData.js
@@ -114,11 +114,13 @@ export default url => FormatView => {
             } = this.props;
 
             if (error) {
-                return (
+                return error === 'bad value' ? (
                     <InvalidFormat
                         format={field.format}
                         value={resource[field.name]}
                     />
+                ) : (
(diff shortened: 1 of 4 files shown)


src/app/js/formats/injectData.js
```diff
@@ -108,11 +109,17 @@ export default url => FormatView => {
...
 
             if (error) {
-                return (
+                return error === 'bad value' ? (
+                    <InvalidFormat
+                        format={field.format}
+                        value={resource[field.name]}
+                    />
+                ) : (
                     <p style={styles.message}>{polyglot.t('chart_error')}</p>
```
Kmaschta, line 123: Please use `AdminOnlyAlert` component to hide this error message when logged as a user

Kmaschta (changes requested):

src/app/js/formats/injectData.js
ThieryMichel: No, I think the user would want to know he has connexion problem

Kmaschta (approved):

PR #745, opened 2018-10-16
Webpack v4

ThieryMichel:
new entry in config/development-dist.js: jsHost to allow to take the jes from the webpack dev server in development(http://localhost:8080). default to ''.


src/api/controller/front.js
```diff
@@ -212,29 +217,13 @@ if (config.userAuth) {
...
-    app.use(mount('/', serve(path.resolve(__dirname, '../../build'))));
-    app.use(mount('/', serve(path.resolve(__dirname, '../../app/custom'))));
-}
+app.use(
+    route.get('/admin', async ctx => {
+        ctx.body = adminIndexHtml;
+    }),
+);
+
+app.use(mount('/', serve(path.resolve(__dirname, '../../build'))));
```
Kmaschta, line 226: Does this mount should be available on dev?

package.json
```diff
@@ -199,12 +202,16 @@
...
     "pre-commit": "1.2.2",
     "prettier": "1.8.2",
     "react-addons-test-utils": "15.6.2",
     "react-hot-loader": "3.1.2",
     "sinon": "4.1.2",
-    "webpack-dev-server": "2.9.4"
+    "webpack": "4.20.2",
+    "webpack-cli": "3.1.2",
+    "webpack-dev-server": "3.1.9",
+    "uglifyjs-webpack-plugin": "2.0.1"
```
Kmaschta: Uglify is already bundled in webpack 4.
Please, remove this dependency.

Kmaschta (changes requested):
Can you remove `react-hot-loader` from webpack ?
Does webpack dev server hot reload already?

src/api/controller/front.js
ThieryMichel: Since it does not interfer, and could be useful to check the build, might as well leave it, No ?

ThieryMichel:
react-hot-loader removed

Kmaschta (approved):

PR #746, opened 2018-10-17
Upgrade to Node 10 and Babel latest

Kmaschta:
## Docker container user

I made the docker containers work with the host user.
The node modules will no longer be installed as `root`.

After this PR, you might change the owner of your files from root to your users.
To do so, you can run the following command on the repo's root folder:

```bash
sudo chown `id -nu`:`id -ng` . -R
```

## Removed pre-commit hook
I removed the pre-commit hook, because didn't work and was badly configured.
You might need to remove the pre commit hook manually:

```bash
rm .git/hooks/pre-commit
```


babel.config.js
```diff
@@ -0,0 +1,40 @@
...
+                },
+                useBuiltIns: 'entry',
+            },
+        ],
+        '@babel/preset-react',
+    ],
+    plugins: [
+        '@babel/plugin-proposal-class-properties',
+        '@babel/plugin-proposal-object-rest-spread',
+        '@babel/plugin-syntax-dynamic-import',
```
Kmaschta, line 17: Allow to use `import()` with Node

package.json
```diff
@@ -180,6 +176,9 @@
     "xml-writer": "1.7.0"
   },
   "devDependencies": {
+    "@babel/core": "7.1.2",
+    "babel-core": "7.0.0-bridge.0",
```
Kmaschta: Needed for Jest.
See https://jestjs.io/docs/en/getting-started.html#using-babel

ThieryMichel (approved):

PR #758, opened 2018-10-22
Nav menu

ThieryMichel:
The drawer will be reintroduced in the next PR


src/app/js/public/NavBar.js
```diff
@@ -0,0 +1,180 @@
...
+    logout,
+    p: polyglot,
+}) => {
+    let img;
+    return (
+        <nav className={css(styles.container)}>
+            <img
+                className={css(styles.icon)}
+                src="http://localhost:8080/favicon.ico"
+                ref={el => (img = el)}
```
Kmaschta, line 111: This is no longer the way to create ref in React.
Please use `React.createRef`

Kmaschta (changes requested):

src/app/js/public/NavBar.js
ThieryMichel: Does not work anymore when I do that...

Kmaschta: Did you try with a real `function` and not an arrow function?

Kmaschta (approved):

PR #773, opened 2018-10-29
Themable menu

ThieryMichel:


config.json
```diff
@@ -83,5 +83,82 @@
...
+      },
+      {
+        "label": {
+                "en": "About",
+                "fr": "A propos"
+        },
+        "icon": "/about.svg",
+        "position": "top",
+        "role": "custom",
+        "link": "/about.html"
```
Kmaschta: This page should not be used by default, right ?

src/api/controller/front.js
```diff
@@ -146,9 +146,8 @@ export const getRenderingData = async (
 
 const handleRender = async (ctx, next) => {
     const { url, headers } = ctx.request;
-
     if (
-        url.match(/[^\\]*\.(\w+)$/) ||
+        (url.match(/[^\\]*\.(\w+)$/) && !url.match(/[^\\]*\.html$/)) ||
```
Kmaschta, line 152: Why just not use `url.startsWith('/')` and `url.endsWith('.html')` .. ?

src/api/controller/index.js
```diff
@@ -18,6 +21,28 @@ if (simulatedLatency) {
...
 app.use(mount('/api', api));
 
+const scriptRegEx = new RegExp('<script.*?( src=".*")?.*?>.*?</script>', 'gm');
+
+export const getScriptsFromHtml = html =>
+    (html.match(scriptRegEx) || [])
+        .map(script => {
+            const src = script.match(/<script.*?src="(.*?)".*?>/);
+            return src && src[1];
+        })
```
Kmaschta, line 31: Please also check that the scripts are relative or on the same host.

src/app/custom/about.html
```diff
@@ -0,0 +1,4 @@
+<div>
```
Kmaschta: Are you sure you want to add this file by default ?

src/app/custom/about.js
```diff
@@ -0,0 +1 @@
+alert('It works');
```
Kmaschta: Hm.

src/app/js/public/MenuItem.js
```diff
@@ -0,0 +1,297 @@
...
+    handleGraphItemClick,
+    toggleSearch,
+    canBeSearched,
+    role,
+    logout,
+}) => {
+    const label = config.label[polyglot.currentLocale];
+    const icon = getIcon(config.icon);
+
+    switch (config.role) {
```
Kmaschta: You should use `toLowerCase` here, just to be sure

src/app/js/public/MenuItem.js
```diff
@@ -0,0 +1,297 @@
...
+                    activeClassName={css(styles.active)}
+                >
+                    {icon}
+                    {label}
+                </NavLink>
+            );
+        }
+        default:
+            console.error(
+                `Unknow role: ${config.role} menu item: ${JSON.stringify(
```
Kmaschta: You just have to use `custom` as the default, no ?
I would avoid to get an error message for nothing.

src/app/js/public/NavBar.js
```diff
@@ -24,7 +14,15 @@ import Drawer from './Drawer';
 import Search from './search/Search';
 import GraphSummary from './graph/GraphSummary';
 import Favicon from './Favicon';
-import theme from '../theme';
+import jsonConfig from '../../../../config.json';
```
Kmaschta: Are you importing the WHOLE config, with its secrets and stuff on the frontend build?

src/app/js/public/Routes.js
```diff
@@ -8,6 +8,12 @@ import Resource from './resource/Resource';
 import Login from '../user/Login';
 import GraphPage from './graph/GraphPage';
 import NavBar from './NavBar';
+import jsonConfig from '../../../../config.json';
+import getCustomPage from './getCustomPage';
+
+const customRoutes = jsonConfig.front.menu.filter(
+    ({ role, link }) => (role !== 'custom' ? false : link.startsWith('/')),
```
Kmaschta: `custom` or no role at all?

src/app/js/public/getCustomPage.js
```diff
@@ -0,0 +1,52 @@
...
+import Helmet from 'react-helmet';
+import translate from 'redux-polyglot/translate';
+import fetch from 'fetch-with-proxy';
+
+import { polyglot as polyglotPropTypes } from '../propTypes';
+
+export class CustomPage extends Component {
+    state = {};
+    UNSAFE_componentWillMount() {
+        fetch(`/customPage${this.props.link}`)
```
Kmaschta: Why the `$` ? Why not just use a query param or just `/customPage/page`?

src/app/js/public/getCustomPage.js
```diff
@@ -0,0 +1,52 @@
...
+            .catch(error => {
+                console.error(error);
+                this.setState({ error: true });
+            });
+    }
+    render() {
+        const { p: polyglot } = this.props;
+        const { html, scripts, error } = this.state;
+        if (error) {
+            return <div>{polyglot.t('page_not_found')}</div>;
```
Kmaschta: Perhaps, try a redirect to the home ?
Just an idea.

src/app/js/public/getCustomPage.js
```diff
@@ -0,0 +1,52 @@
...
+        }
+        if (!html) {
+            return null;
```
Kmaschta: Idem, error or redirect to the home

src/app/js/public/getCustomPage.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+CustomPage.propTypes = {
+    link: PropTypes.string.isRequired,
+    p: polyglotPropTypes.isRequired,
+};
+
+const getCustomPage = link =>
+    translate(function CustomPageForLink() {
+        return <CustomPage link={link} />;
```
Kmaschta: This is not the best implementation that you can do.

Kmaschta (changes requested):

src/app/js/public/getCustomPage.js
ThieryMichel: :trollface:

src/app/js/public/index.js
```diff
@@ -30,16 +31,24 @@ const store = configureStore(
...
+        </MuiThemeProvider>
+    </Provider>,
     document.getElementById('root'),
 );
 
 // Hot Module Replacement API
 if (module.hot) {
     module.hot.accept('../Root', () => {
-        const NewRoot = require('../Root').default; // eslint-disable-line
+        const NewRoutes = require('./Routes').default; // eslint-disable-line
```
Kmaschta: Are you sure that we still need this `module.hot` condition now we have a webpack dev server?

src/app/translations.tsv
```diff
@@ -423,3 +423,4 @@
 "host.isbn"	"ISBN"	"ISBN"
 "host.eisbn"	"E-ISBN"	"E-ISBN"
 "host.title"	"Title"	"Titre"
+"page_not_found"	"Page not found"	"Page non trouvée"
```
Kmaschta: Is this translation still used.. ?

Kmaschta (changes requested):

src/app/js/public/MenuItem.js
ThieryMichel: nope, i think it's better to be strict for config

Kmaschta (approved):

PR #775, opened 2018-10-31
Reset graph filters when switching graphs

Kmaschta:


src/app/js/public/Toolbar.js
```diff
@@ -1,13 +1,22 @@
...
+import PropTypes from 'prop-types';
 
 import FacetList from './facet/FacetList';
 import Filter from './dataset/Filter';
 
-export const ToolbarComponent = () => (
+export const Toolbar = ({ name }) => (
     <div>
-        <Filter />
+        <Filter key={name} />
```
Kmaschta, line 9: Protip: By adding the graph page name as the key of this component, it is resetted each time the page change (eg. from `/graph` to `/graph/abcd`)

ThieryMichel: :+1:

ThieryMichel (approved):

PR #783, opened 2018-11-05
added new transformer to avoid to catch wrong string value (Iike a wrong ISSN)

touv:


src/common/transformers/MASK.spec.js
```diff
@@ -0,0 +1,10 @@
+import { mask } from './MASK';
+
+describe.only('MASK', () => {
```
Kmaschta: `.only` found

Kmaschta (changes requested):

src/common/transformers/MASK.spec.js
touv: oups, it's fixed

Kmaschta (approved):

PR #789, opened 2018-11-07
New format streamgraph

youlot:


Kmaschta:
Hey ! Is this PR ready for review or still WIP ?
Just let us know by setting the right label so we know if we can review :)

youlot:
I forgot to set the label sorry :)

src/app/js/formats/streamgraph/StreamAdmin.js
```diff
@@ -0,0 +1,48 @@
...
+    //    onChange={this.setHeightValue}
+    ///>
+    //<TextField
+    //    floatingLabelText="color"
+    //    onChange={console.log("color")}
+    ///>
+
+    render() {
+        return (
+            <div style={styles.container}>
```
Kmaschta: This component renders nothing.
So, it's better to delete this file and not expose `AdminComponent` at all.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
+import React, { PureComponent } from 'react';
+import { connect } from 'react-redux';
+import compose from 'recompose/compose';
+import * as d3 from 'd3';
+
+import { getCssToString, zoomFunction, distinctColors, transformDataIntoMapArray, randomId, getMinMaxValue } from './utils';
+import injectData from '../injectData';
+import exportableToPng from '../exportableToPng';
+
+const styles = {
```
Kmaschta: Please use the `aphrodite` library to manage your styles.
This way, the users will be able to custom it if needed.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
+import React, { PureComponent } from 'react';
+import { connect } from 'react-redux';
+import compose from 'recompose/compose';
+import * as d3 from 'd3';
+
+import { getCssToString, zoomFunction, distinctColors, transformDataIntoMapArray, randomId, getMinMaxValue } from './utils';
```
Kmaschta: Did you run a linter? (we use prettier)

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+        this.yAxisScale;
+        this.xAxis;
+        this.yAxisL;
+        this.yAxisR;
+        this.gx;
+        this.gyr;
+        this.gyl;
+        this.streams;
+
+        //this.draft = draft.bind(this);
```
Kmaschta: Commented code found. Please delete any commented code.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+        this.zoomFunction = zoomFunction.bind(this);
+    }
+
+    setGraph() {
+        let [valuesObjectsArray, valuesArray, minDate, maxDate, nameList] = transformDataIntoMapArray(this.props.formatData);
+        const divContainerWidth = document.getElementById(this.divContainer).clientWidth;
```
Kmaschta: This don't use the `document` API directly from a React component.
To manipulate DOM elements, you can use the [React refs API](https://reactjs.org/docs/refs-and-the-dom.html).

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+class Streamgraph extends PureComponent {
+
+    constructor(props) {
+        super(props);
+        this.state = {
+            width: 800,
+            height: 300
+        }
+
+        this.uniqueId = randomId(10);
```
Kmaschta: By using the React refs API, you can access to the DOM elements of your components without having to setting a uniq ID.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+    setGraph() {
+        let [valuesObjectsArray, valuesArray, minDate, maxDate, nameList] = transformDataIntoMapArray(this.props.formatData);
+        const divContainerWidth = document.getElementById(this.divContainer).clientWidth;
+        let { height } = this.state;
+        let width = divContainerWidth;
+        const svgHeight = height;
+        const svgWidth = width;
+
+        const d3DivContainerId = "d3DivContainer" + this.uniqueId;
+        const divContainer = d3.select(this.refs.divContainer)
```
Kmaschta: Is the variable `this.refs` initialized? I don't see it.
Please, use the React refs API, it would be far more readable.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+        const d3DivContainer = divContainer.append("div")
+            .attr("id", d3DivContainerId);
+        const svgViewport = d3.select(this.refs.anchor);
+        let svgContainerDiv = d3.select(this.refs.svgContainer);
+        let componentContext = this;
+
+        // ===========================================================================================
+        // Set all the variables
```
Kmaschta: You are forced to use this kind of comments because the `setGraph` function is HUGE.
Please, split it into multiple well named functions so that you don't have to write this kind of comments.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+        let componentContext = this;
+
+        // ===========================================================================================
+        // Set all the variables 
+        // ===========================================================================================
+
+        svgContainerDiv.attr("width", divContainerWidth);
+        let layersNumber = valuesObjectsArray.length;
+
+        // stack the datas
```
Kmaschta: This comment is not really useful, the code is clear enough.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+
+        // append y axis to the left of the chart
+        this.gyl = innerSpace.append("g")
+            .attr("class", "y axis")
+            .call(this.yAxisL);
+
+        // color set
+        const z = distinctColors(layersNumber);
+        //var z = d3.interpolateRainbow;
+        //var z = d3.interpolateCool;
```
Kmaschta: Commented code again

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+            legendItemContainer.append("text")
+                .attr("style", getCssToString(styles.legentItemText, styles))
+                .attr("class", "legentItemText")
+                .text(element[0]);
+        }
+
+        // ===========================================================================================
+        // Mouse events - svgViewport events
+        // 
+        // Can't split in different blocks because .on("") events
```
Kmaschta: Again, this comment is here because the function is too large. Please split it into multiple ones.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+                </svg>
+            </div>
+        );
+    }
+};
+
+const mapStateToProps = (state, { formatData }) => {
+    //console.log("mapStateToProps - state:");
+    //console.log(state);
+    return ({});
```
Kmaschta: If you `mapStateToProps` returns nothing, you don't have to write it.
Please delete this function and remove the `connect` below.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,410 @@
...
+};
+
+export default compose(injectData(), connect(mapStateToProps), exportableToPng)(
+    Streamgraph,
+);
```
Kmaschta: Please add an empty line at the end of every of your files.

src/app/js/formats/streamgraph/index.js
```diff
@@ -0,0 +1,9 @@
+import Component from './Streamgraph';
+import AdminComponent from './StreamAdmin';
+import DefaultFormat from '../DefaultFormat';
+
+export default {
+    Component,
+    AdminComponent,
```
Kmaschta: AdminComponent is empty, you can remove it.

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,204 @@
+
+import differenceBy from 'lodash/differenceBy';
+import * as d3 from 'd3';
+
+// ===========================================================================================
+// getCssToString get the css js object and return it as string
```
Kmaschta: Please, remove this kind of comment.
They take too much space and if the function is small enough and well named, we don't need them.

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,204 @@
...
+// ===========================================================================================
+
+export function getCssToString(cssToChange, styles) {
+    const cssString = Object.keys(styles.legend).map(key => {
```
Kmaschta: You don't have to translate styles into CSS: aphrodite can do that for you.
Just apply the correct class names.

Kmaschta (changes requested):
I may have several feedbacks to your PR, but let's stick with the more important ones:

- Please, don't push commented code
- Use a linter (prettier is already installed, you can configure your IDE to use it)
- Write some unit tests, even some basic ones

I know the time is flying but we would keep a code clean and maintainable.
Also, I would recommend to iterate faster on your PR (even if it's WIP code).
That way, we'll be able to review early and avoid you a lot of rewriting.

Finally, we're available on slack if you have any question or need help!

ThieryMichel (changes requested):

src/app/js/formats/streamgraph/Streamgraph.js
youlot: I use this styles in d3 method like this : `.attr("style", getCssToString(styles.tooltip, styles))`
I did this way because d3 needs the css as a string. I had a look to aphrodite and it doesn't seems to have a function to get the css as a string, it only want to inject css through the dom which is not good for this kind of use with d3. 

What do you think about it ? Did I miss something ?

(For now I can setup the css with aphrodite and use it with my method getCssToString, it works)

src/app/js/formats/streamgraph/Streamgraph.js
youlot: I installed prettier in visual studio code, thanks !

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,204 @@
...
+        colors.push(hslToHex(hue, 90, 50));
+    }
+    return colors;
+}
+
+// ===========================================================================================
+// hslToHex : Convert hsl values color to hex color
+// ===========================================================================================
+
+export function hslToHex(h, s, l) {
```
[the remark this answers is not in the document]
youlot: because when I was looking for a method to have different colors around the palette I find this function so I just copy paste it ! But yes, could use d3-color

src/app/js/formats/streamgraph/Streamgraph.js
Kmaschta: Indeed, aphrodite generates class names.

I see that you're editing the `style` attribute to set inline CSS to it.
IMHO, you can do it the same, but with the `class` attribute in order to set class names to it.

For example:
```js
import { StyleSheet, css } from 'aphrodite/no-important';

const styles = StyleSheet.create({
    tooltip: { /* tooltip style */ },
});

// [...]
selector.attr('class', css(styles.tooltip));
```

youlot: Absolutely yes, just didn't think to it.. thanks !

src/app/js/formats/streamgraph/Streamgraph.js
youlot: For the refs, I tryed to set up using the style `<div ref={this.myRef} />` but d3js cannot append on it through a `d3.select(this.myRef).append("div")`. I have any idea why, searched a lot on the web and only find examples with the `<div ref="myRef" />` style.. Ideas ?

Kmaschta: @youlot You have to set one ref per element you want to reach:

```jsx
class MyComponent extends Component {
    contructor(props) {
        super(props);
        this.containerRef = React.createRef();
        this.svgRef = React.createRef();
        this.anchorRef = React.createRef();
    }

    render() {
        const { width, height } = this.state;
        return (
            <div id="divContainer" ref="this.containerRef" style={styles.divContainer}>
                <svg id="svgContainer" ref="this.svgRef" width={width} height={height}>
                    <g id="anchor" ref="this.anchorRef" />
                </svg>
            </div>
        );
    }
}
```

Then, [as explained on the doc](https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs), in order to access the DOM element pointed by the ref, you have to write:

```js
console.log(this.containerRef.current);
// So, the following should work:
d3.select(this.containerRef.current).append("div")
```

If needed, here is the API of a DOM element (the `current` part of the ref): https://developer.mozilla.org/en-US/docs/Web/API/element

youlot: @Kmaschta Yes I tried this method as described in the documentation, one ref per element e.g: `this.containerRef`, initialized in constructor with `React.createRef();` and assigned to an html element using `ref={this.containerRef}`. This works well for using the ref from react but I have any idea why d3js cannot use it, when I try a select() using this ref it just doesn't work as if the ref was broken.

I finally figured it out by using callbacks refs : ` ref={divContainer => this.divContainer = divContainer}` doing this way the ref is good from react side and d3js can access to this ref ! I don't understand yet why this method works while the previous one doesn't but.. it works !

Kmaschta: The refs callback is the legacy way of doing so. It's not optimal and will be deprecated soon.

Did you try with the current key ? `this.containerRef.current` when you selected?

youlot: Are you sure the callback refs will be deprecated ? I haven't seen anything about it in the [documentation](https://reactjs.org/docs/refs-and-the-dom.html). I have seen that the string ref will be removed in one of the future releases and that's why they recommend using the callback pattern or the createRef api.

I tried the .current way in association with the string ref assignment and it didn't worked but doing the assignment this way : `ref={this.divContainer}` works with the .current ! I will commit this usage because I think that's more appropriate.

Kmaschta: I'm not fully certain that it'll be deprecated, but indeed, the second option is best! Good job on that :1st_place_medal:

youlot: Okay, thanks for your help ! :)

src/app/js/formats/streamgraph/index.js
```diff
@@ -0,0 +1,9 @@
...
+    EditionComponent: DefaultFormat.EditionComponent,
```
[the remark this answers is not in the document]
youlot: I found how to add an icon, will do it soon :)

Kmaschta:
Instead of merge master to your branch, we'd prefer that you rebase it.

Be sure to have the both branches up-to-date:

```bash
git checkout master && git pull master
git checkout [branch] && git pull [branch]
```
Then you can rebase:

```bash
git rebase master
# or `git rebase -i master` (and let only the commit of your PR in the list)
```
At this point, you might have some conflicts, but GitHub warns you about them if there are some.

Then, when the rebase is done, you have to push force on your branch to apply the rebase.

src/app/js/formats/streamgraph/StreamAdmin.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
```
Kmaschta: As explained in previous comment, please delete this file.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+    this.state = {
+      width: 800,
+      height: 300
+    };
+
+    this.divContainer = React.createRef();
+    this.svgContainer = React.createRef();
+    this.anchor = React.createRef();
+
+    this.xAxisScale;
```
Kmaschta: What's the point to these declarations?
You don't have to declare them in order to this "this.x" later.

Also, it's really a bad habit to use `this.x` to store values in JS.
Please delete these declarations.

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,159 @@
+import differenceBy from "lodash/differenceBy";
+import * as d3 from "d3";
+
+export function zoomFunction() {
+  // create new scale ojects based on event
+  let new_xScale = d3.event.transform.rescaleX(this.xAxisScale);
```
Kmaschta: Instead of binding `this` to this function, you have two solutions:
- Pass all what you needs in argument
- Declare this function as a method of the component class

I'd prefer that you use the second option.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+    this.gx;
+    this.gyr;
+    this.gyl;
+    this.streams;
+
+    this.hoveredKey;
+    this.hoveredValue;
+    this.mouseIsOverStream = false;
+
+    this.zoomFunction = zoomFunction.bind(this);
```
Kmaschta: Please declare the zoom function as a method of this class.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+      .scaleExtent([1, 32])
+      .translateExtent([[0, 0], [width, height]])
+      .extent([[0, 0], [width, height]])
+      .on("zoom", this.zoomFunction);
+
+    const innerSpace = svgViewport
+      .append("g")
+      .attr("width", width)
+      .attr("class", "inner_space")
+      .attr("transform", "translate(" + margin.left + "," + margin.top + ")");
```
Kmaschta: You can use template literal here, it's more readable.

```js
`translate(${margin.left}, ${margin.top})`
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+      .attr("clip-path", "url(#mask)")
+      .call(zoom);
+
+    graphZone
+      .append("rect")
+      .attr("width", "100%")
+      .attr("height", "100%")
+      .attr("opacity", 0);
+
+    return [innerSpace, graphZone];
```
Kmaschta: It's very pythonic!
But prefer to return an object, in this way you can use object destructuring later.

```js
return { innerSpace, graphZone };
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+  }
+
+  stackDatas(nameList, valuesArray) {
+    let stackMethod = d3
```
Kmaschta: I have a suggestion for you:

```js
stackDatas(nameList, valuesArray) {
    if (valuesArray.length <= 0) {
        return null;
    }

    const stackMethod = d3
      .stack()
      .keys(nameList)
      .order(d3.stackOrderNone)
      .offset(d3.stackOffsetWiggle);

      return stackMethod(valuesArray);
}
```

Let me know if you find it more readable :)

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+  createAndSetAxis(
+    minDate,
+    maxDate,
+    minValue,
+    maxValue,
+    width,
+    height,
+    innerSpace
+  ) {
+    minDate = new Date(String(minDate));
```
Kmaschta: You are mutating the minDate here.
This is a really bad idea because it will update it in the whole scope and the parent of this functions will have a transformed value.

Please declare a new variable for that.

```js
createAndSetAxis(min, max, ...) {
    const minDate = new Date(min); // etc
}
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+    maxDate = new Date(String(maxDate));
+
+    this.xAxisScale = d3
```
Kmaschta: It would be reaaaaally nice if you avoid to declare things in `this`.
React doesn't like that (because of the whole rendering lifecycle).

If you can't, it will cause performance issues (but if it works it's not that bad)

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+      .call(this.yAxisR);
+
+    this.gyl = innerSpace
+      .append("g")
+      .attr("class", "y axis")
+      .call(this.yAxisL);
+  }
+
+  createAndSetStreams(layersNumber, graphZone, stackedData, nameList) {
+    let colorNameList = []; // e.g.: [Thematique,#FFFFFF]
```
Kmaschta: You are forced to use a comment to explain what is inside of that value.
But an object would do this for you!

```js
const colorNameList = [];
colorNameList.push({ name, color });
```

In general, we prefer to use objects instead of lists.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+        return this.yAxisScale(d[1]);
+      });
+
+    this.streams = graphZone
+      .selectAll("path")
+      .data(stackedData)
+      .enter()
+      .append("path")
+      .attr("d", d => {
+        if (d) {
```
Kmaschta: A small protip to increase readability:

```js
.attr('d', d => !!d ? area(d) : [])
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+          return area(d);
+        } else {
+          return [];
+        }
+      })
+      .attr("name", (d, i) => {
+        let name = nameList[i];
```
Kmaschta: Please, always use `const` by default and only let if you are mutating the value.
But in general, you don't have to mutate anything.

```js
colorNameList.push({ name: nameList[i], color: z[i] });

return name;
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+        let color = z[i];
+        colorNameList.push([name, color]);
+        return name;
+      })
+      .attr("opacity", 1)
+      .attr("fill", (d, i) => {
```
Kmaschta:
```js
.attr("fill", (d, i) => z[i])
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+      .insert("div", "#svgContainer")
+      .attr("class", `remove ${css(styles.tooltip)}`)
+      .attr("id", "tooltip")
+      .style("position", "absolute")
+      .style("z-index", "2")
+      .style("visibility", "hidden")
+      .style("top", "4px")
+      .style("left", "12px");
+
+    return [tooltip, vertical];
```
Kmaschta: Again, please use an object

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+  }
+
+  createAndSetTheLegend(d3DivContainer, colorNameList) {
+    const legendView = d3DivContainer
+      .append("div")
+      .attr("id", "legend")
+      .attr("class", `${css(styles.legend)}`);
+
+    for (let index = colorNameList.length - 1; index > 0; index--) {
```
Kmaschta: Protip ! We no longer use for loops, but instead:

```js
colorNameList.forEach((item, index) => {
    // Do the flop
});
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+
+  setMouseMoveAndOverStreams(tooltip) {
+    let componentContext = this;
+
+    this.streams
+      .on("mousemove", function(d, i) {
+        let mousex = d3.mouse(this)[0];
+        let date = componentContext.xAxisScale.invert(mousex);
+        componentContext.mouseIsOverStream = true;
+        this.hoveredKey = d3.select(this).attr("name");
```
Kmaschta: Please stop putting things in this.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+
+        for (let elem of d) {
+          if (elem.data.date.getFullYear() == date.getFullYear()) {
+            this.hoveredValue = elem.data[this.hoveredKey];
```
Kmaschta: For loops are usually not readable.
You can replace this whole loop by one line:

```js
const hoveredValue = d
    .find(elem => elem.data.date.getFullYear() === date.getFullYear())
    .data[hoveredKey];
```
PS: In JS `==` is not safe, please use `===`

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+    // Can't split events in different blocks because .on("") events
+    // definition erase the previous ones
+    this.setViewportEvents(svgViewport, vertical);
+    this.setMouseMoveAndOverStreams(tooltip);
+    this.setMouseOutStreams(tooltip);
+  }
+
+  setGraph() {
+    let [
+      valuesObjectsArray,
```
Kmaschta: Please use object destructuring here, it works the same.

```js
const {
     valuesObjectsArray,
     valuesArray,
      minDate,
      maxDate,
      nameList,
} = transformDataIntoMapArray(this.props.formatData);
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+      valuesArray,
+      minDate,
+      maxDate,
+      nameList
+    ] = transformDataIntoMapArray(this.props.formatData);
+
+    let svgWidth = this.divContainer.current.clientWidth;
```
Kmaschta: const over let

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+
+    let { height } = this.state;
```
Kmaschta: This will help you not declaring multiple time the same variable:
```js
const { height: svgHeight } = this.state;
```

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,476 @@
...
+    let svgHeight = height;
+    const margin = { top: 60, right: 40, bottom: 50, left: 60 };
+    let width = svgWidth - margin.left - margin.right;
+    height = svgHeight - margin.top - margin.bottom;
```
Kmaschta: Please, stop mutate things.

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,159 @@
...
+            };
+            valuesObjectsArray.push(currentItem);
+          }
+
+          currentItem.values.push({
+            date: new Date(elem.source),
+            value: elem.weight
+          });
+
+          if (parseInt(elem.source) < dateMin || dateMin === -42) {
```
Kmaschta: parseInt is unsafe and needs to have a second argument to set the base. `parseInt(elem.source, 10)`

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,159 @@
...
+    // display the datas at the center when only one x value
+    if (dateMin == dateMax) {
+      dateMin--;
+      dateMax++;
+    }
+  }
+
+  let namesList = [];
+  for (let element of valuesObjectsArray) {
+    namesList.push(element.name);
```
Kmaschta: Please don't use for.

```js
const namesList = values.map(value => value.name);
```

Kmaschta (changes requested):
I still have some feedbacks !

Please, use `const` by default, instead of let.
Also, there is still some comment that had not been addressed.
And finally, I'm not sure your linter is configured as the config (or not used?)

src/app/js/formats/streamgraph/StreamAdmin.js
youlot: @Kmaschta @ThieryMichel  I fully understand and agree with the idea to not put useless code on the repo but I would like to let this file because once the pr is accepted I will directly work on the admin part. So if I delete this file and all the links to it I will have to set it up just after and I think it's counterproductive.

What do you think ?

Kmaschta:
> @Kmaschta okay for the rebase ! Why do you prefer the usage of this technique over the merge ?

Objectively, there is no "best" way to manage out-dated branches, that be merge or rebase.
It's just an habit we have and it's better to have the same practices on the same project.
We prefer rebase because, it doesn't let "merge commits" everywhere :)

src/app/js/formats/streamgraph/StreamAdmin.js
Kmaschta: Sure, it's understandable!
But we consider each commit on master might be deployed: actually, it's the case, we deploy to production every few PR merged.
Hence, we won't deploy some dead code.

That said, you can have stashed code on your local env without commiting it.

TLDR: WIP code should not end to prod.

src/app/js/formats/streamgraph/Streamgraph.js
Kmaschta: I'm somewhat OK with that fact that it doesn't bother React.

But, do you agree with me that this code has no effect and can be removed?

```js
this.xAxisScale;
this.yAxisScale;
this.xAxis;
this.yAxisL;
this.yAxisR;
this.gx;
this.gyr;
this.gyl;
this.streams;
this.hoveredKey;
this.hoveredValue;
```

youlot: yes absolutely ! I put it there for the understanding of the code but if you think that's useless I delete it, no problems.

src/app/js/formats/streamgraph/Streamgraph.js
youlot: Done !

src/app/js/formats/streamgraph/Streamgraph.js
youlot: This was a mistake ! I fixed it, thanks.

src/app/js/formats/streamgraph/Streamgraph.js
youlot: yes, python habits here, that's fixed.

src/app/js/formats/streamgraph/StreamAdmin.js
youlot: Okay I understand, I will do the necessary.

src/app/js/formats/streamgraph/utils.js
youlot: I don't understand why but when I move the declaration of this function in the class through a method it fails, the method have some problems with this.xAxisScale which is empty. I lost a lot of time on it so I will let the function in this state for now.

youlot:
I made all the possible changes, can you check it and validate the pr ? I need it to continue the work on the others features.

src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -0,0 +1,497 @@
...
+  legendButton: {
+    height: 15,
+    width: 15
+  }
+});
+
+class Streamgraph extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
```
Kmaschta: I see no `this.setState` in this component. This, these values (width and height) are constants and don't needs to be store in state.

src/app/js/formats/streamgraph/StreamgraphIcon.svg
```diff
@@ -0,0 +1,2 @@
...l-1.3-0.5l0,0l0,0l-1.3-0.5l-1.3-0.5L86,83.6l0,0l0,0l-1.3-0.5l-1.3-0.5l-1.3-0.5l-1.3-0.4l-1.2-0.4l-1.2-0.4l-1.2-0.4l-1.2-0.4  l-1.2-0.4l-1.2-0.3l-1.2-0.3l-1.2-0.3L70,78.6l-1.1-0.2l-1.1-0.2l-1.1-0.1l0,0l0,0l-1.1-0.1l-1.1,0l-1.1,0l-1,0l-1,0.1l-1.1,0.2  l-1.1,0.2l-1,0.2L57,79.2l-1.1,0.3l-1,0.4l-1.1,0.4l-1.1,0.4l-1.1,0.5l-1.1,0.5l-1,0.5l-1.1,0.5l0,0l0,0l-1.1,0.5l0,0l0,0l-1.1,0.5  l-1.1,0.5l0,0l0,0l-1.1,0.4l0,0l0,0l-1,0.4l-0.1,0l-0.1,0l-1,0.3l0,0l0,0l-1,0.2l-0.1,0l-0.1,0l-1,0.2l-0.1,0l-0.1,0l-1,0l-0.1,0  l-0.1,0l-1-0.1l-0.1,0l-0.1,0l-1-0.2l-0.1,0l-0.1,0l-1-0.4l-0.1,0l-0.1,0l-1-0.5l-0.1-0.1l-0.1-0.1l-1-0.7l-0.1-0.1L32,83.7l-1-0.8  l-0.1,0l-0.1,0l-1-1l0,0l0,0l-1-1.1l0,0l0,0l-1-1.2l0,0l0,0l-1-1.3l-1-1.3L25,75.3l-0.9-1.2L23.2,73l-0.8-0.9l-0.7-0.6l-0.4-0.2  l-0.1,0L21,71.3l-0.5,0.4l-0.7,1.1L19,74.5l-0.9,2.3l-0.9,2.8l-0.9,3.2l-0.4,1.6l0.5-1l0.1-0.1l0.1-0.1l0.9-1.2l0.2-0.2l0.2-0.2  l0.9-0.7l0.3-0.2l0.3-0.1l0.9-0.3l0.5-0.1l0.5,0.1l0.9,0.2l0.3,0.1l0.2,0.1l0.9,0.5l0.2,0.1l0.2,0.1l0.9,0.9l0.1,0.1l0.1,0.1  l0.9,1.1l0,0.1l0,0.1l0.9,1.2l1,1.3l0,0l0,0l0.9,1.4l0.9,1.3l0.9,1.2l0.9,1.1l0.9,1l0.9,0.8l0.8,0.7l0.8,0.5l0.8,0.4l0.8,0.3  l0.8,0.2l0.7,0l0.8-0.1l0.9-0.1l0.9-0.3l0.9-0.3l0.9-0.4l1-0.5L44.6,92.8z"/>
```
Kmaschta: Are you importing the `StreamgraphIcon.svg` somewhere ?
If no, you can delete it.

src/app/js/formats/streamgraph/index.js
```diff
@@ -0,0 +1,9 @@
+import Component from './Streamgraph';
+import DefaultFormat from '../DefaultFormat';
+import Icon from './StreamgraphIcon';
+
+export default {
+    ...DefaultFormat,
+    Component,
+    Icon,
+};
```
Kmaschta, line 9: Please let a line at the end of the file.

Kmaschta (changes requested):
The PR is mostly OK for me!

There are still some minor things:

- I think you used the wrong linter because you have two spaces instead of 4.
- In `streamgraph/utils.js`, you can use D3 helpers (like d3-color)
- `transformDataIntoMapArray` is totally unreadable by a human.

Also, you can rebase from master to fix the tests (we fixed some things)

src/app/js/formats/streamgraph/utils.js
```diff
@@ -0,0 +1,175 @@
...
+          maxValue = value[1];
+        }
+      }
+    }
+  }
+
+  return { minValue, maxValue };
+}
+
+export function cutStr(str) {
```
Kmaschta, line 160: Let me help you with this function.

```js
// Naming is important: please don't use shurtnames (like str)
// It's always better to use full meaningful names for variables, it helps the reader a lot
export function limitWords = (str, maxWordNumber = 3) => {
    const words = str.split(' ');
    const sentence = words.filter((word, index) => index <= maxWordNumber).join(' ');

    return words.length > maxWordNumber
        ? `${sentence} ...`
        : sentence;
};
```

youlot: Your solution doesn't produce the same result.

src/app/js/formats/streamgraph/index.js
youlot: :+1:

src/app/js/formats/streamgraph/StreamgraphIcon.svg
youlot: it's not, I took the bubble chart as example which do the same. I though you was putting the svg to keep a trace of the origin file but no problems, will delete it.

src/app/js/formats/streamgraph/utils.js
Kmaschta: I even haven't understood your code, then.
The goal isn't to get the first three words and add `...` if there is a rest?

youlot: You missed the if conditions.. look [here](https://repl.it/@taspai/BoldAntiqueMemoryallocator)

Kmaschta: I get a 404 Not Found :/

Kmaschta (approved):

PR #792, opened 2018-11-08
fix custom link

ThieryMichel:


Kmaschta (commented):
I suggest to add a `exact` prop to the Route, this way it will fail into a 404 without an additional request.

PR #794, opened 2018-11-08
add more drawer width for windows scroll bar

ThieryMichel:
https://trello.com/c/pi81CwwW/80-affichage-des-graph-sur-2-colonnes-que-sur-linux-%C3%A0-voir-pour-windows


Kmaschta (commented):
Please, can you change the graph icons / squares size instead?
Thank you very much!

PR #799, opened 2018-11-09
fix facet sort by count

ThieryMichel:


Kmaschta (changes requested):
Can you add a test somewhere, so that it doesn't happen next time?

src/app/js/public/facet/FacetValueList.js
```diff
@@ -39,27 +39,29 @@ const onPageChange = (changeFacetValue, filter, name) => (
...
+) => (_, filter) =>
     changeFacetValue({
         name,
         currentPage,
         perPage,
         filter,
     });
 
-const onInvertChange = (invertFacet, name) => (_, inverted) =>
+export const onInvertChange = (invertFacet, name) => (_, inverted) =>
```
Kmaschta: Why do you export these functions?

Kmaschta (approved):

Kmaschta (approved):

PR #804, opened 2018-11-12
Use ReactDOM.hydrate instead of Render

Kmaschta:
https://reactjs.org/docs/react-dom.html#hydrate

## Before

image

## After

image


ThieryMichel:
e2e test fail

ThieryMichel (dismissed):

PR #811, opened 2018-11-15
Hide other issues

ThieryMichel:


src/app/js/formats/istexSummary/FetchFold.js
```diff
@@ -46,6 +47,55 @@ const circularProgress = (
     />
 );
 
+class SkipFold extends Component {
```
Kmaschta: Can you move this component into its own file please?

src/app/js/formats/istexSummary/FetchFold.js
```diff
@@ -46,6 +47,55 @@ const circularProgress = (
...
+
+        if (isLoading) {
+            return circularProgress;
+        }
+
+        if (!data.hits.length) {
+            return null;
+        }
+
+        return this.props.children({ ...this.props, data });
```
Kmaschta: You can use object destructuring here :)

Kmaschta (changes requested):

Kmaschta (approved):

PR #821, opened 2018-11-21
update args with format name to avoid incoherent state

ThieryMichel:


Kmaschta (approved):

Kmaschta (commented):
Why do we still need the state? Can we remove the state from FormatEdition?

ThieryMichel:
We need it to have default args in case of obsolete model loaded

PR #843, opened 2018-11-27
Update format streamgraph

youlot:


touv (changes requested):
tests failed, can you fix it ?

youlot:
I made some fixes for the tests but now they failed on the make install : 

`Pulling api (node:10)...
10: Pulling from library/node
ERROR: no matching manifest for linux/amd64 in the manifest list entries
make: *** [install-npm-dependencies] Error 1
The command "make install" failed and exited with 2 during .`

I don't know what to do because those logs doesn't really help..

touv:
Maybe, we are in this case :  https://github.com/docker-library/official-images/issues/3835

youlot:
Yes, this is maybe the solution and it explains why the tests have passed now

PR #845, opened 2018-12-04
Streamgraph admin colors

youlot:


src/app/js/formats/streamgraph/Streamgraph.js
```diff
@@ -576,14 +576,14 @@ class Streamgraph extends PureComponent {
...
                         position: 'absolute',
                         top: `${height / 2 - 30}px`,
-                        left: `${margin.left + width / 2 - 130}px`,
+                        left: `${margin.left + width / 2 - 275}px`,
                         color: 'white',
                     }}
                 >
                     <h4>
-                        Use mouse scroll to zoom and
+                        Utilisez la molette de la souris pour zoomer
```
touv: Why don't you use the translation system ?

like this
```
 {polyglot.t('Use mouse scroll to zoom and')}
```
You can find a use case example of polyglot in this file https://github.com/Inist-CNRS/lodex/blob/master/src/app/js/formats/istex/IstexView.js

touv (changes requested):

src/app/js/formats/streamgraph/Streamgraph.js
touv: ...and all translations are in this file https://raw.githubusercontent.com/Inist-CNRS/lodex/master/src/app/translations.tsv

touv (approved):

PR #846, opened 2018-12-06
Use ezs-lodex statements instead of local statements

parmentf:

PR #862, opened 2018-12-21
Revert "Revert "Use ezs-lodex statements instead of local statements""

parmentf:
Reverts Inist-CNRS/lodex#859 (and restore #846).


parmentf:
That's what I wanted to fix (even when the tests in Travis are green):
image
My last commit did it.

The failing tests in travis took 11 min 21 sec but seemed to be successful.
Now, the succeeding tests in travis take 20 min 35 sec (maybe the duration of the tests is the best indicator of their success).

PR #879, opened 2019-02-20
Add the max value parameter in admin panel

youlot:


touv (changes requested):
Why don't you use the RoutineParamsAdmin component ?
You can use it like the other graphs. It's more consistent.
https://github.com/Inist-CNRS/lodex/blob/b43989d1bee8f7bf7d307fe0e905a10c90e8f725/src/app/js/formats/network/NetworkAdmin.js#L71-L75


https://github.com/Inist-CNRS/lodex/blob/master/src/app/js/formats/network/NetworkAdmin.js#L71-L75

youlot:
Because on the Hierarchy graph we don't need the minValue, maxValue, orderBy parameters.

touv:
I think this function is  is useless : 
```
    setMaxSize = (_, maxSizeValue) => {
        this.setParams({ maxSize: maxSizeValue });
    };
```

youlot:
Yes, that was a mistake

touv (approved):

PR #884, opened 2019-02-26
Hierarchy trucated labels

youlot:


touv (changes requested):
cutText function doesn't support utf8 encoding. In this case, I think that it would be better  to use a existing package like "cli-trunacte"  (UTF8 support, already testing, widely used)
https://www.npmjs.com/package/cli-truncate

touv:
@youlot you should commit and push the package-lock.json (see travis tests)

touv (approved):

PR #919, opened 2019-04-24
Use external scripts for exporters

touv:
Make exporters use lodex-extented scripts when they exist.

First example: `extended-nquads.ini`


parmentf:
Similar to src/api/controller/api/upload.js (for loaders) and run.js (for routines) but in export.js

parmentf:
Strange: the export filename is not what is given in `Content-Disposition` header.
It's always `export.` label (instead of extension).

parmentf:
[Corriger l'extension des fichiers exportés](https://trello.com/c/tFIL4skl/248-corriger-lextension-des-fichiers-export%C3%A9s)

parmentf:
Tests still fail because of `src/api/e2e/ssr.spec.js` which tries to get exporters from the ancient Javascript-written exporters (through `src/api/exporters/index.js`).

These tests have to use `src/api/controller/api/export.js` `getExporters`.
And this `getExporters` should not return empty exporters, as it does now.

parmentf:
[externaliser les exports NQuads](https://trello.com/c/guYJ4Csm/245-externaliser-les-exports-nquads)

PR #922, opened 2019-06-14
harmonize color input

CorentinSardo:
### Step 1 :
Every input should be with text
### TODO :
Bubble & Network


touv (changes requested):
I think you should read this blog :
 https://kentcdodds.com/blog/please-dont-commit-commented-out-code

CorentinSardo:
Not really sure it's fine testing (it doesn't use mapStateToProps), but tests are now valid and the same checks are done.
In fact, colors in text mode aren't tested yet in the app, so I had to try to do it by myself.

CorentinSardo:
Shall we factorize the code color input code in this PR or in the next one ?

src/app/js/formats/title/TitleAdmin.js
```diff
@@ -21,7 +21,7 @@ const styles = {
 
 export const defaultArgs = {
     level: 1,
-    textColor: '#000',
+    textColor: '#818181',
```
parmentf: The idea is neat, but I would have used a constant (defined in an external file, perhaps within src/app/js/formats).

src/app/js/formats/hierarchy/Hierarchy.js
```diff
@@ -300,7 +284,12 @@ class Hierarchy extends PureComponent {
                         d.parent.y
                     );
                 })
-                .attr('stroke', color);
+                .attr(
+                    'stroke',
+                    /(^#[0-9A-F]{6}$)|(^#[0-9A-F]{3}$)/i.test(color)
```
parmentf: This limits the color to hexadecimal codes.
Maybe we should create a function `isValidColor` and use it as is now.
That way, we could improve the function later.
And the function could be common to all color parameters.

src/app/js/formats/bubbleChart/BubbleView.js
```diff
@@ -48,7 +48,13 @@ export const BubbleView = ({ data, diameter, colorSet }) => (
                     y={y}
                     name={key}
                     value={value}
-                    color={colorSet[index % colorSet.length]}
+                    color={
+                        /(^#[0-9A-F]{6}$)|(^#[0-9A-F]{3}$)/i.test(
```
parmentf: Same code here.
Definitely, time for a function (see previous comment).

parmentf (changes requested):

parmentf:
Oh. It seems it was already present in original code, but `eslint(react/prop-types)` rule is triggered within `Hierarchy.js`, and likely in other formats too.

One day, we'll have to improve that.

src/app/js/formats/bubbleChart/BubbleView.js
CorentinSardo: Definitely better that way, external file [src/app/js/formats/colorUtils.js](https://github.com/Inist-CNRS/lodex/blob/293798c30299e4ce54b8d594d9c4c31034cc2977/src/app/js/formats/colorUtils.js) added

PR #923, opened 2019-06-24
ColorPicker in his own Component

CorentinSardo:
V1 testable in the Radar graph, feature will be improved in this PR.


CorentinSardo:
Code needs to be improved, but this version works better than the first.

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+    constructor(props) {
+        super(props);
+        isFirstLoading = true; // to initialize values on admin UI opening
+        const colorsArray = this.getColorsArray();
+
+        this.state = {
+            colors: [{ color: colorsArray[0] }],
+        };
+
+        let numberOfPickers = this.getColorsArray().length;
```
parmentf:
```suggestion
        let numberOfPickers = colorsArray.length;
```

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+        for (let index = 1; index < numberOfPickers; index++) {
+            this.state = {
+                colors: [...this.state.colors, { color: colorsArray[index] }],
+            };
+        }
```
parmentf: I would have used:

```js
        this.state = {
            colors: colorsArray.map(color => ({ color }))
        };
```

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+    }
+
+    constructor(props) {
+        super(props);
+        isFirstLoading = true; // to initialize values on admin UI opening
+        const colorsArray = this.getColorsArray();
+
+        this.state = {
+            colors: [{ color: colorsArray[0] }],
+        };
```
parmentf: Why make a case here for the first color, and a second for the remaing colors?

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -92,6 +169,15 @@ class RadarChartAdmin extends Component {
                     style={styles.input}
                     value={colors}
                 />
+                {this.createUI()}
+                <br />
```
parmentf: Can't you manage breaking spaces using CSS margins?
I know it can be complicated, though.

parmentf (changes requested):

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+        let numberOfPickers = this.getColorsArray().length;
+        let res = '';
+        for (var i = 0; i < numberOfPickers; i++) {
+            if (i == 0) {
+                res += this.state.colors[i].color;
+            } else {
+                res += ' ' + this.state.colors[i].color;
+            }
+        }
+        this.setColors;
```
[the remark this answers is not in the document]
CorentinSardo: Oh yes it does nothing, this one should be deleted

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+    setColors = (_, colors) => {
+        updateAdminArgs('colors', colors, this.props);
+    };
+
+    setColorsWithColorPicker = () => {
+        updateAdminArgs('colors', this.getColorsFromPicker(), this.props);
+    };
+
+    getColorsFromPicker() {
+        let numberOfPickers = this.getColorsArray().length;
```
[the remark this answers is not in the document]
CorentinSardo: Yeah the length was an issue !
I fixed it by adjusting the array size during my UI updating to match the correct size : 
For now I got this better version, because of the length of `state.colors` problem :
```javascript
            [...this.state.colors] = this.state.colors.slice(
                0,
                colorsArray.length,
            );
```

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+        let res = '';
+        for (var i = 0; i < numberOfPickers; i++) {
+            if (i == 0) {
+                res += this.state.colors[i].color;
+            } else {
+                res += ' ' + this.state.colors[i].color;
+            }
+        }
+        this.setColors;
+        return res;
```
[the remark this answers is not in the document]
CorentinSardo: With the fix of `state.colors.length`, it works fine, nice !

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+    setColorsWithColorPicker = () => {
+        updateAdminArgs('colors', this.getColorsFromPicker(), this.props);
+    };
+
+    getColorsFromPicker() {
+        let numberOfPickers = this.getColorsArray().length;
+        let res = '';
+        for (var i = 0; i < numberOfPickers; i++) {
+            if (i == 0) {
+                res += this.state.colors[i].color;
```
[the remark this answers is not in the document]
CorentinSardo: After some testing, the `state` one contains the colorPickers data and the `props` one the updated value in the `TextField` containing the color codes

Both seem to be necessary to make everything work !

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
CorentinSardo: Oh I was stuck for a long time with this assignment (the only way I have found was the 0 + loop), but this mapping works fine ! Thanks !

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
CorentinSardo: (See upper issue)

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
CorentinSardo: Oh yeah I now see how it's made, I'll to go for that if needed :+1:

src/app/js/formats/distributionChart/global-radar-chart/RadarChartAdmin.js
```diff
@@ -75,10 +76,86 @@ class RadarChartAdmin extends Component {
...
+        return res;
+    }
+
+    getColorsArray() {
+        return this.props.args.colors.split(' ');
+    }
+
+    constructor(props) {
+        super(props);
+        isFirstLoading = true; // to initialize values on admin UI opening
```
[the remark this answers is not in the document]
CorentinSardo: Well, thanks for highlighting the lifecycle thing, since it's really important and wasn't familiar with it !

However the problem deals specifically with the `render` method, so I guess I will go for the property or the state solution

CorentinSardo:
[See for update version in a component](https://github.com/Inist-CNRS/lodex/pull/924)

src/app/js/formats/bubbleChart/BubbleAdmin.js
```diff
@@ -50,20 +51,32 @@ class BubbleAdmin extends Component {
         args: defaultArgs,
     };
 
+    constructor(props) {
+        super(props);
+        this.handleColorsChange = this.handleColorsChange.bind(this);
+        this.state = {
+            colors:
+                this.props.args.colors != null
```
parmentf: I see that comparison very seldomly now.

I would have written

```js
this.props.args.colors || colorUtils.MULTICHROMATIC_DEFAULT_COLORSET,
```

Sure, there may be some cases where it does not work well, but which ones?

parmentf: Also, `colorUtils.MULTICHROMATIC_DEFAULT_COLORSET` should have the same value as `defaultArgs.colors`, so why not use the latter as a default value (advantage: only one location to define default value).

parmentf: Oh, wait. That's what you use line 91: `this.state.colors || defaultArgs.colors`.

src/app/js/formats/bubbleChart/BubbleAdmin.js
```diff
@@ -50,20 +51,32 @@ class BubbleAdmin extends Component {
         args: defaultArgs,
     };
 
+    constructor(props) {
+        super(props);
+        this.handleColorsChange = this.handleColorsChange.bind(this);
```
parmentf: I don't get the reason why you need to bind `this.handleColorsChange`, but not `this.setDiameter`, nor `this.setParams`.

src/app/js/formats/bubbleChart/BubbleAdmin.js
```diff
@@ -50,20 +51,32 @@ class BubbleAdmin extends Component {
...
 
     setDiameter = (_, diameter) => {
         updateAdminArgs('diameter', diameter, this.props);
     };
 
-    setColors = (_, colors) => {
+    handleColorsChange(colors) {
         updateAdminArgs('colors', colors, this.props);
-    };
+        this.setState({ colors });
```
parmentf: Why save the `colors` in the state?
That ColorPickerParamsAdmin component should deal with `colors`, shouldn't it?

src/app/js/formats/colorUtils.js
```diff
@@ -1,6 +1,6 @@
-export const MONOCHROMATIC_DEFAULT_COLORSET = '#818181';
+export const MONOCHROMATIC_DEFAULT_COLORSET = '#2b83ba';
 export const MULTICHROMATIC_DEFAULT_COLORSET =
-    '#1D1A31 #4D2D52 #9A4C95 #F08CAE #C1A5A9';
+    '#d7191c #fdae61 #ffffbf #abdda4 #2b83ba';
 export const MULTICHROMATIC_DEFAULT_COLORSET_STREAMGRAPH =
     '#E6194B #3CB44B #FFE119 #4363D8 #F58231 #911EB4 #46F0F0 #F032E6 #BCF60C #FABEBE #008080 #E6BEFF #9A6324 #FFFAC8 #800000 #AAFFC3 #808000 #FFD8B1 #000075 #808080 #FFFFFF #000000';
```
parmentf: prettier seems to want an empty at the end of a file.

src/app/js/formats/bubbleChart/BubbleAdmin.js
CorentinSardo: > Binding methods helps ensure that the second snippet works the same way as the first one.
([React docs](https://reactjs.org/docs/faq-functions.html#why-is-binding-necessary-at-all))

It is necessary to link the parent and the new child Component `ColorPickerParamsAdmin` !

src/app/js/formats/bubbleChart/BubbleAdmin.js
CorentinSardo: Oh yeah that is clearer !

src/app/js/formats/bubbleChart/BubbleAdmin.js
CorentinSardo: > [...] we want these two inputs to be in sync with each other. [...]
In React, sharing state is accomplished by moving it up to the closest common ancestor of the components that need it. This is called “lifting state up”. We will remove the local state from the [Child] and move it into the [Parent] instead.
([React docs](https://reactjs.org/docs/lifting-state-up.html#lifting-state-up))

In our case it is needed in order to sync the colorPickers and the TextField !
It seems to be a the classic pattern to do that with React

src/app/js/formats/bubbleChart/BubbleAdmin.js
parmentf: It's used in `onColorsChange`, and not in `onChange`, like in the other cases.

OK

src/app/js/formats/bubbleChart/BubbleAdmin.js
CorentinSardo: Just made a change to stop using `props.colors` in the Component : I managed to use it effectively as read only data.
This way, all the changes are inside of the Component, with `state.colors`, and this `setState` is now useless in the parent ! 😄

parmentf (approved):

PR #931, opened 2019-09-09
Use new @ezs packages

parmentf:
Because we gather all ezs packages (`ezs` and `ezs-*`) within one `@ezs` npm organization.

See  <https://www.npmjs.com/org/ezs>
See <https://github.com/Inist-CNRS/ezs>

Ideally, this PR should be merged before any other development.


parmentf:
Don't forget that some external scripts (like those in [lodex-extended](https://github.com/Inist-CNRS/lodex-extended/tree/v3)) can use `@ezs` packages that are not explicitly imported in lodex's code.

That's why you can't remove any non-used `@ezs` package, say `@ezs/lodex`, for example.

Luwangel (approved):

PR #971, opened 2019-10-18
remove CSS rules for the look & feel

touv:
and make the default theme more basic & lean


Luwangel (changes requested):
The default style is broken. It should be at least usable (see screenshot).

Sélection_019

Luwangel:
Fixed in branch https://github.com/Inist-CNRS/lodex/pull/1007

PR #984, opened 2019-10-28
Load loaders dynamically

mchaffotte:
[Trello Card #5](https://trello.com/c/1XCtwe9p/5-impossible-dajouter-dynamiquement-un-nouveau-loader-alors-que-lon-peut-ajouter-une-routine)

It's possible to add a routine dynamically, but not a loader. The objective is to inject new loaders without creating a new version of loadex.

## Todo

- [x] Add loader dynamically
- [x] Add tests

### DIFF
--- a/src/app/js/admin/loader/sagas/loadLoaders.js
+++ b/src/app/js/admin/loader/sagas/loadLoaders.js
@@ -1,0 +1,31 @@
+import { call, takeEvery, select, put } from 'redux-saga/effects';
+
+import {
+    PRE_LOAD_LOADERS,
+    loadLoaders,
+    loadLoadersError,
+    loadLoadersSuccess,
+} from '..';
+import { fromUser } from '../../../sharedSelectors';
+import { fromLoaders } from '../../selectors';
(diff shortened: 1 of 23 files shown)


src/app/js/admin/loader/sagas/loadLoaders.js
```diff
@@ -0,0 +1,31 @@
...
+    loadLoaders,
+    loadLoadersError,
+    loadLoadersSuccess,
+} from '..';
+import { fromUser } from '../../../sharedSelectors';
+import { fromLoaders } from '../../selectors';
+import fetchSaga from '../../../lib/sagas/fetchSaga';
+
+export function* handleLoadLoaders() {
+    if (yield select(fromLoaders.areLoadersLoaded)) {
```
Luwangel: I don't like this syntax due to its readability. I think it's better to write :

```
const loaderAlreadyLoaded = yield ...;

if (loaderAlreadyLoaded) {
    return;
}
```

Luwangel (changes requested):

Luwangel (approved):

PR #97, opened 2017-02-22
[RFR] Adding transformer

ThieryMichel:

PR #1016, opened 2019-12-03
Dynamic translations

mchaffotte:
[Trello Card #97](https://trello.com/c/EW5FsXEI/97-faire-une-%C3%A9tude-sur-comment-g%C3%A9rer-les-traductions-dynamiquement)

## Todo

- [x] Put the translation file in the custom folder
- [x] Use the same translation file in the app and the admin
- [x] Create a route to return translations (for dynamic calls)
- [x] Fix tests


src/api/controller/front.js
```diff
@@ -68,8 +68,8 @@ const getDefaultInitialState = (token, cookie, locale) => ({
         locale: locale,
         phrases:
             locale === 'fr' || locale === 'fr-FR'
```
parmentf: Why not take the two first letters of the `locale`?
Here is the list of possibilities:

- fr-BE
- fr-CA
- fr-CH
- fr-FR
- fr-LU
- fr-MC

mchaffotte: You are right, I use one solution found in another file but better is to check only the two first letters.

Luwangel (approved):

PR #1034, opened 2019-12-12
Graph - Load graph with filters

mchaffotte:
[Trello Card #90](https://trello.com/c/cDbAiiM7/90-ne-pas-r%C3%A9initialiser-les-facettes-automatiquement-quand-on-quitte-la-page-graphique)

## Todo

- [x] Apply facets when loading the graph

### DIFF
--- a/src/app/js/formats/sagas.js
+++ b/src/app/js/formats/sagas.js
@@ -89,11 +89,15 @@ export function* loadFormatData(name, url, queryString) {
     yield put(loadFormatDataSuccess({ name, data: response }));
 }
 
+const isUrl = text => {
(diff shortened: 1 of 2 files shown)


src/app/js/formats/sagas.js
```diff
@@ -89,11 +89,15 @@ export function* loadFormatData(name, url, queryString) {
     yield put(loadFormatDataSuccess({ name, data: response }));
 }
 
+const isUrl = text => {
```
Luwangel: You could use (and maybe improve) the `isURL` method in `uris.js`

Luwangel (changes requested):

Luwangel (approved):

PR #1046, opened 2019-12-23
Update lodex extended and remove sitemap from default exporters

touv:


config.json
```diff
@@ -57,15 +57,13 @@
     ],
     "exporters": [
         "csv",
+        "tsv",
```
Luwangel, line 60: Why did you remove "sitemap" from exporters?

Luwangel (approved):

PR #1077, opened 2020-06-22
Update vega and vega lite dependencies

AlasDiablo:


AlasDiablo:
A problem remain with the Area Charts not rendering correctly

src/app/js/formats/vega-lite/VegaLiteView.js
```diff
@@ -20,7 +20,7 @@ class VegaLiteView extends Component {
         let spec;
 
         try {
-            spec = JSON.parse(specTemplate);
+            spec = this.retroCompatibilityCheck(JSON.parse(specTemplate), data);
```
AlasDiablo: The `data` parameter can be remove for now, but is possible to be used for fixing area charts

AlasDiablo:
I found the problem for the AreaCharts.
Theres are some deprecated field in the Area Charts template use by use, this can be only fixed by editing the template.
Suggestion of new basic template :
```json
{
   "width":600,
   "height":200,
   "mark":"area",
   "selection":{
      "dataset":{
         "type":"multi",
         "fields":[
            "target"
         ],
         "bind":"legend"
      }
   },
   "encoding":{
      "x":{
         "timeUnit":"yearmonth",
         "field":"source",
         "type":"temporal",
         "axis":{
            "domain":false,
            "format":"%Y",
            "tickSize":0
         }
      },
      "y":{
         "aggregate":"sum",
         "field":"weight",
         "type":"quantitative",
         "stack":"center",
         "axis":null
      },
      "color":{
         "field":"target",
         "type":"nominal",
         "scale":{
            "scheme":"category20b"
         }
      },
      "opacity":{
         "condition":{
            "selection":"dataset",
            "value":1
         },
         "value":0.2
      }
   }
}
```

touv (approved):

PR #1079, opened 2020-06-29
add barchart and refactors vage-lite folder

AlasDiablo:


AlasDiablo:
I close this PR because:
+ I break some thing from my previous PR.
+ I need to change some thing with the code architecture.
+ I need to impl the VegaLite spec with a other way (use pre-build models with the code from this PR).

PR #1080, opened 2020-06-30
Change the avalible actions in Vega Embed

AlasDiablo:
Remove from vega embed:
+ View Source
+ View Compiled Vega
+ Open in Vega Editor


src/app/js/formats/vega-lite/component/vega-component/VegaLiteComponent.js
```diff
@@ -0,0 +1,49 @@
...
+                svg: true,
+                png: true,
+            },
+            source: false,
+            compiled: false,
+            editor: false,
+        };
+    }
+
+    let spec = props.spec;
```
touv: use const instead of let, and I think this syntax is more conventional :

const { spec } = props;

see https://evertpot.com/javascript-let-const/ & https://exploringjs.com/es6/ch_destructuring.html

PR #1086, opened 2020-07-24
Attempt to fix a bug with the data manipulation (AsterPlot)

AlasDiablo:


touv:
see https://github.com/Inist-CNRS/lodex/pull/1084#issuecomment-663073676

PR #1110, opened 2020-10-13
Correctif Import Ancien Modele

jdemangeon:
https://trello.com/c/crX5ji1p/25-permettre-de-corriger-un-mod%C3%A8le-qui-nest-plus-compatible-avec-la-nouvelle-version-de-lodex

### DIFF
--- a/cypress/integration/publication.spec.js
+++ b/cypress/integration/publication.spec.js
@@ -82,6 +82,19 @@ describe('Dataset Publication', () => {
                 .should('be.visible');
         });
 
+        it('should allow to fix obsolete imported model before publish dataset', () => {
+            menu.openAdvancedDrawer();
+            menu.goToAdminDashboard();
+            datasetImportPage.importDataset('dataset/simple.csv');
+            datasetImportPage.importModel('model/concat-obsolete.json');
+
+            cy.contains('Errors').click();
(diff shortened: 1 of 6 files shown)


cypress/integration/publication.spec.js
```diff
@@ -82,6 +82,19 @@ describe('Dataset Publication', () => {
...
+            menu.openAdvancedDrawer();
+            menu.goToAdminDashboard();
+            datasetImportPage.importDataset('dataset/simple.csv');
+            datasetImportPage.importModel('model/concat-obsolete.json');
+
+            cy.contains('Errors').click();
+            cy.contains('Operation UNKNOWN-OPERATION does not exist').click();
+
+            datasetImportPage.setOperationTypeInWizard('BOOLEAN');
+            datasetImportPage.publish();
```
JulienMattiussi, line 95: It's stange that this test ends with a publish instruction, and not with a testing instruction.
Is the publish needed ? 
And if it does, isn't it testable ?

jdemangeon: I can test it's "publishable", but publish will fail if it's not. So, it act like a test on "publishable".
You known what I mean?

JulienMattiussi: Forget it, it's ok

JulienMattiussi (approved):

PR #1114, opened 2020-10-15
Keep navigation links (previous and next resource)

JulienMattiussi:
## Issue [24](https://trello.com/c/lH6cGzK5)

Navigation to next and previous resources are lost when the page is reloaded

## Solution 

Keep this data in sessionStorage


jdemangeon (approved):
I don't understand what you do exactly but it seems right :rofl: 
Is it already a session storage system setup?

JulienMattiussi:
Yes, there was already a session storage system to keep user informations. I just reused it for keeping search informations.
It's using "redux-localstorage" (old but looking stable anyway)

PR #1143, opened 2020-11-24
Import file drag n drop (Part 3)

JulienMattiussi:
Follow #1142 

Last part of import file using drag n drop

**Mode file**

image

**Switch en mode url**

image

### DIFF
--- a/src/app/js/admin/upload/index.js
+++ b/src/app/js/admin/upload/index.js
@@ -18,9 +18,11 @@ export const closeUpload = createAction(CLOSE_UPLOAD);
 export const changeUploadUrl = createAction(CHANGE_UPLOAD_URL);
 export const changeLoaderName = createAction(CHANGE_LOADER_NAME);
 
-const validateUrl = url =>
-    console.log({ url }) ||
-    (url && (url.startsWith('http://') || url.startsWith('https://')));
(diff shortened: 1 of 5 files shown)


src/app/js/admin/upload/index.js
```diff
@@ -18,9 +18,11 @@ export const closeUpload = createAction(CLOSE_UPLOAD);
 export const changeUploadUrl = createAction(CHANGE_UPLOAD_URL);
 export const changeLoaderName = createAction(CHANGE_LOADER_NAME);
 
-const validateUrl = url =>
-    console.log({ url }) ||
-    (url && (url.startsWith('http://') || url.startsWith('https://')));
+const validateUrl = url => {
+    if (url && (url.startsWith('http://') || url.startsWith('https://')))
```
jdemangeon: Just return the condition will works, "if" is useless here

jdemangeon (changes requested):

src/app/js/admin/upload/index.js
JulienMattiussi: There was a bug saying the return type is not always boolean with this syntax

jdemangeon (approved):

PR #1151, opened 2020-11-27
Fix count and labels in statistics footer

JulienMattiussi:
- Do not pluralize when there only zero or one in count
- Use correct labels
- Fix count with filter in display page

image

image

image


JulienMattiussi:
Obsolete

PR #1160, opened 2020-12-08
use /tmp/upload  in production mode

touv:


Dockerfile.spec
```diff
@@ -12,7 +12,7 @@ RUN npm install --production && \
 COPY ./src /app/src
 COPY ./config /app/config
 COPY ./config.json ./babel.config.js ./jest.config.js ./jsconfig.json ./typings.json /app/
-RUN mkdir /app/upload
+RUN mkdir /tmp/upload
```
touv, line 15: I don't know if it' really useful

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
             no_proxy: 'localhost,127.0.0.1,'
             BABEL_DISABLE_CACHE: 1
         volumes:
-            - ./upload:/app/upload
+            - ./upload:/tmp/upload
```
touv: doesn't work or not ?

PR #1173, opened 2020-12-17
Fix fields count

JulienMattiussi:
Don't count subresources field in resources pages


jdemangeon (approved):
Good, but does it works in the subresource edit page?

PR #1183, opened 2021-01-15
Move Action Buttons And Delete Ready Model

jdemangeon:

PR #1184, opened 2021-01-15
Add subresource to model export (follows #1183)

jdemangeon:
I've completly removed old "field import". Should I keep it as a fallback when the imported file is a "json" and not "tar" ? We should discuss it during next daily meeting


jdemangeon:
Rebased

JulienMattiussi (approved):

PR #1203, opened 2021-02-03
Fix zip import

JulienMattiussi:
Fix zip import errors when The zip file contains more than 1000 items

Règle les cartes

- https://trello.com/c/1eItBpV7
- https://trello.com/c/e8WiROMP
- https://trello.com/c/wXtvQpHZ


src/api/services/initDatasetUri.js
```diff
@@ -1,11 +1,13 @@
 import getDocumentTransformer from '../../common/getDocumentTransformer';
 import { URI_FIELD_NAME } from '../../common/uris';
+import progress from './progress';
+import { INITIALIZING_URI } from '../../common/progressStatus';
 
 const initDatasetUriFactory = ctx =>
     async function initDatasetUri() {
         const count = await ctx.dataset.countWithoutUri();
+        progress.start(INITIALIZING_URI, count);
```
jdemangeon, line 9: Why should'nt you put it before "countWithoutUri" ?

src/api/models/dataset.js
```diff
@@ -41,16 +41,19 @@ export default db => {
...
-            items.map(item => ({
-                updateOne: {
-                    filter: getFilter(item),
-                    update: { $set: item },
-                    upsert,
-                },
-            })),
-        );
+    collection.bulkUpdate = async (items, getFilter, upsert = false) => {
+        if (items.length) {
```
jdemangeon, line 45: A "bulkWrite" with empty array should work, no?

src/api/controller/api/upload.js
```diff
@@ -72,14 +72,12 @@ export const uploadFile = ctx => async loaderName => {
     const parsedStream = parseStream(mergedStream);
     try {
         await ctx.saveParsedStream(parsedStream, ctx.initDatasetUri);
+        progress.finish();
```
jdemangeon, line 75: Why didn't you finish "progress" after "clearChunks" like before ?

JulienMattiussi: The goal is to clearChunks even if progress hadn't finished correctly
Because previously, when something went wrong, chunks were staying, and were taken twice at the next upload

src/api/models/dataset.js
JulienMattiussi: It look like it fails in mongo. But we are using 2.2 and bulkWrite is tagged as a 3.2 feature, so strange behaviour are not surprising

src/api/services/initDatasetUri.js
JulienMattiussi: To put the count inside the progress as a target

jdemangeon: My bad --"

src/api/controller/api/upload.js
jdemangeon: ok :+1:

src/api/models/dataset.js
jdemangeon: OK :+1:

jdemangeon (approved):

PR #1213, opened 2021-02-05
Use wildcard to list index keys

JulienMattiussi:
The createIndex function in mongodb dont accept "Index keys" document too heavy
When there was more than 85 searchable fields in the model, it failed.

https://trello.com/c/luYMwqcO


jdemangeon:
Current MongoDB version can't handle index exclusion, so every field is indexed for the moment... We choose to make fields not searchable to avoid error when index count exceed 85 fields.

PR #1228, opened 2021-03-01
Fix legend display

JulienMattiussi:
- Limit legend to same scope fields
- Remove legends from graphic list
- Fix graphLink 'voir les détails' display for home and for legends

### DIFF
--- a/src/common/scope.js
+++ b/src/common/scope.js
@@ -2,6 +2,20 @@ export const SCOPE_DATASET = 'dataset';
 export const SCOPE_GRAPHIC = 'graphic';
 export const SCOPE_COLLECTION = 'collection';
 export const SCOPE_DOCUMENT = 'document';
+
+export const sameScope = (scopeA, scopeB) => {
+    if (scopeA === scopeB) {
+        return true;
+    }
+    if (
+        (scopeA === SCOPE_DOCUMENT || scopeA === SCOPE_COLLECTION) &&
+        (scopeB === SCOPE_DOCUMENT || scopeB === SCOPE_COLLECTION)
+    ) {
+        return true;
+    }
+    return false;
+};
+
 export const SCOPES = [
     SCOPE_DATASET,
     SCOPE_GRAPHIC,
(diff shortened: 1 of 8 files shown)


src/common/scope.js
```diff
@@ -2,6 +2,20 @@ export const SCOPE_DATASET = 'dataset';
 export const SCOPE_GRAPHIC = 'graphic';
 export const SCOPE_COLLECTION = 'collection';
 export const SCOPE_DOCUMENT = 'document';
+
+export const sameScope = (scopeA, scopeB) => {
```
jdemangeon: This function return true in every case no :grin: ?

If a & b  = SCOPE_DOCUMENT
if a & b = SCOPE_COLLECTION
if a = SCOPE_COLLECTION & b = SCOPE_DOCUMENT 
if b = SCOPE_COLLECTION & a = SCOPE_DOCUMENT

jdemangeon (changes requested):
I don't understand all you've developed, but I trust you!

src/common/scope.js
JulienMattiussi: For these cases, yes
but when a = SCOPE_DATASET and b = SCOPE_GRAPHIC, then it's false

JulienMattiussi:
> I don't understand all you've developed, but I trust you!

Basically, before, Lodex was checking that the field was a graphic to set somes specific displays.
But after, I replaced it by a check on the Scope : The field is in the graphic page
It's wrong because all the fields in the graphic page are not graphics
So now, I changed the tests again to fit with reality (legends are not graphics field, and some cloned field can be here and should be considered as graphics)

src/common/scope.js
jdemangeon: OK, so can you use includes instead, and rename the function because it not only test for equality

JulienMattiussi: Use fantastic functional syntax from JulienD's big brain

jdemangeon (approved):

PR #1245, opened 2021-04-27
Update refbibs format

touv:


touv (commented):
the tests failed, a "make install" and the associated commit are missing

PR #1246, opened 2021-04-30
Update resources grid format to be using it inside sub-resources more easily

AlasDiablo:
- [x] Custom summary size
- [x] Css fix
- [ ] `See more` issues *(linked to the routine)* Inist-CNRS/lodex-extended#46
- [x] Open in a new tab

### DIFF
--- a/src/app/js/formats/resources-grid/ResourcesGridView.js
+++ b/src/app/js/formats/resources-grid/ResourcesGridView.js
@@ -61,6 +61,7 @@ class ResourcesGridView extends Component {
         total: PropTypes.number.isRequired,
         pageSize: PropTypes.number.isRequired,
         summarySize: PropTypes.number.isRequired,
+        openInNewTab: PropTypes.bool.isRequired,
         spaceWidth: PropTypes.string.isRequired,
         filterFormatData: PropTypes.func.isRequired,
         allowToLoadMore: PropTypes.bool.isRequired,
@@ -117,6 +118,7 @@ class ResourcesGridView extends Component {
                                 <LodexResource
                                     {...entry}
                                     summarySize={this.props.summarySize}
+                                    openInNewTab={this.props.openInNewTab}
                                 />
                             </div>
                         </li>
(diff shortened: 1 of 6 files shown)


src/app/js/formats/resources-grid/ResourcesGridView.js
```diff
@@ -113,7 +115,11 @@ class ResourcesGridView extends Component {
                             className={styles.item}
                         >
                             <div className={styles.content}>
-                                <LodexResource {...entry} />
+                                <LodexResource
+                                    {...entry}
+                                    summarySize={this.props.summarySize}
+                                    openInNewTab={this.props.openInNewTab}
+                                />
```
AlasDiablo: Better way to do this ?

src/app/js/formats/shared/LodexResource.js
```diff
@@ -39,7 +39,28 @@ const styles = stylesToClassname(
...
+    let url = props.url;
+    let title = props.title;
+    let summary = props.summary;
+    let summarySize = props.summarySize;
+    let openInNewTab = props.openInNewTab;
+
+    if (summarySize > 0) {
+        let currentSize = 0;
+        let finalOutput = '';
+        for (const substr of summary.split(' ')) {
```
jj618: You should put summary.split in a "const" before the "for" to avoid spliting at each iteration.

AlasDiablo: with a forEeach call after the split this can work also ?
```js
summary.split(' ').forEach(e => {/* code */})
```

jj618: Maybe a regex for the split? split(/\s+/) to take care of multi spaces, tab and other stuffs

src/app/js/formats/shared/LodexResource.js
```diff
@@ -39,7 +39,28 @@ const styles = stylesToClassname(
...
+            finalOutput += substr;
+            currentSize += substr.length;
+            if (currentSize >= summarySize) {
+                finalOutput += '…';
+                break;
+            } else finalOutput += ' ';
```
jj618: No need for "else" statement in this case

jj618 (approved):

PR #1258, opened 2021-06-28
feat: (re)add test via github actions

AlasDiablo:


AlasDiablo:
Wow, i think i have make a missing take in the action trigger, is started the test 2 time...

parmentf (approved):
I thought it would have been more complex than that.
Nice job, Pierre!

PR #1276, opened 2021-10-04
24: Remove export button from subresources

arimet:
remove-export-btn


jdemangeon (changes requested):
Top! Can you add a no-regression test?

jdemangeon (approved):

PR #1283, opened 2021-10-06
19: Remove add field for user

arimet:


jdemangeon (changes requested):
You miss to delete the AddField Component that is not used anywhere anymore

jdemangeon (approved):

PR #1284, opened 2021-10-06
17: Remove verifyUri popup for fixing publish

arimet:


jdemangeon (changes requested):
Does nbInvalidUri redux state attribute still used? If not, you miss to delete it

jdemangeon (approved):

PR #1285, opened 2021-10-08
Added an information popup about the data merge before importing a new dataset

arimet:
modalInfo

### DIFF
--- a/cypress/integration/publication.spec.js
+++ b/cypress/integration/publication.spec.js
@@ -31,6 +31,26 @@ describe('Dataset Publication', () => {
                 .eq(1)
                 .should('contains.text', ['2', 'Row 2', 'Test 2'].join(''));
         });
+
+        it('should display a information popup when adding a second dataset', () => {
+            menu.openAdvancedDrawer();
+            menu.goToAdminDashboard();
+            cy.wait(300);
+            datasetImportPage.importDataset('dataset/simple.csv');
+            cy.wait(500);
+
+            cy.get('.sidebar')
+                .contains('a', 'Add')
+                .click();
+
+            cy.wait(300);
+            datasetImportPage.importOtherDataset('dataset/simple.csv');
+            cy.wait(500);
+
+            cy.get('tbody tr')
+                .eq(1)
+                .should('contains.text', ['2', 'Row 2', 'Test 2'].join(''));
+        });
     });
 
     describe('Publication', () => {
--- a/cypress/integration/publication.spec.js
+++ b/cypress/integration/publication.spec.js
@@ -22,8 +22,7 @@ describe('Dataset Publication', () => {
             menu.goToAdminDashboard();
             cy.wait(300);
             datasetImportPage.importDataset('dataset/simple.csv');
-            cy.wait(500);
-            cy.get('tbody tr')
+            cy.get('tbody tr', { timeout: 500 })
                 .eq(0)
                 .should('contains.text', ['1', 'Row 1', 'Test 1'].join(''));
 
@@ -37,9 +36,8 @@ describe('Dataset Publication', () => {
             menu.goToAdminDashboard();
             cy.wait(300);
             datasetImportPage.importDataset('dataset/simple.csv');
-            cy.wait(500);
 
-            cy.get('.sidebar')
+            cy.get('.sidebar', { timeout: 500 })
                 .contains('a', 'Add')
                 .click();
 
--- a/cypress/support/datasetImportPage.js
+++ b/cypress/support/datasetImportPage.js
@@ -18,6 +18,20 @@ export const importDataset = (filename, mimeType = 'text/csv') => {
     cy.get('tbody', { timeout: 6000 }).should('exist');
 };
 
+export const importOtherDataset = (filename, mimeType = 'text/csv') => {
+    fillInputWithFixture('input[type=file]', filename, mimeType);
+    cy.wait(300);
+    cy.get('.btn-upload-dataset').click({ force: true });
+    cy.wait(300);
+    cy.get('#confirm-upload').should('be.visible');
+    cy.wait(300);
+    cy.get('.confirm').click({ force: true });
+    cy.get('.progress').should('exist');
+    cy.get('.progress', { timeout: 6000 }).should('not.exist');
+
+    cy.get('tbody', { timeout: 6000 }).should('exist');
+};
+
 export const importMoreDataset = (filename, mimeType = 'text/csv') => {
     cy.get('.sidebar')
         .contains('Add more')
--- a/cypress/support/datasetImportPage.js
+++ b/cypress/support/datasetImportPage.js
@@ -22,13 +22,11 @@ export const importOtherDataset = (filename, mimeType = 'text/csv') => {
     fillInputWithFixture('input[type=file]', filename, mimeType);
     cy.wait(300);
     cy.get('.btn-upload-dataset').click({ force: true });
+    cy.get('#confirm-upload', { timeout: 300 }).should('be.visible');
     cy.wait(300);
-    cy.get('#confirm-upload').should('be.visible');
-    cy.wait(300);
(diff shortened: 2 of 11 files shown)


cypress/support/datasetImportPage.js
```diff
@@ -18,6 +18,20 @@ export const importDataset = (filename, mimeType = 'text/csv') => {
     cy.get('tbody', { timeout: 6000 }).should('exist');
 };
 
+export const importOtherDataset = (filename, mimeType = 'text/csv') => {
+    fillInputWithFixture('input[type=file]', filename, mimeType);
+    cy.wait(300);
```
jdemangeon, line 23: You should wait for element visibility. Using wait is a bad practice

cypress/support/datasetImportPage.js
```diff
@@ -18,6 +18,20 @@ export const importDataset = (filename, mimeType = 'text/csv') => {
...
+    cy.get('.btn-upload-dataset').click({ force: true });
+    cy.wait(300);
+    cy.get('#confirm-upload').should('be.visible');
+    cy.wait(300);
+    cy.get('.confirm').click({ force: true });
+    cy.get('.progress').should('exist');
+    cy.get('.progress', { timeout: 6000 }).should('not.exist');
```
jdemangeon, line 29: Why 6000?

src/app/js/admin/upload/ConfirmUpload.js
```diff
@@ -0,0 +1,85 @@
...
+import { polyglot as polyglotPropTypes } from '../../propTypes';
+import { fromUpload } from '../selectors';
+
+const styles = {
+    container: {
+        paddingBottom: '1rem',
+    },
+};
+
+export const ConfirmUploadComponent = ({
```
jdemangeon: Can you rename this component? Because we don't known that it's a popup when reading "ConfirmUpload"

src/app/js/admin/upload/ConfirmUpload.js
```diff
@@ -0,0 +1,85 @@
...
+                <div style={styles.container} id="confirm-upload">
+                    {polyglot.t('info_publish_desc')}
+                </div>
+            </DialogContent>
+            <DialogActions>{actions}</DialogActions>
+        </Dialog>
+    );
+};
+
+ConfirmUploadComponent.propTypes = {
```
jdemangeon: You miss isOpenPopup proptype. Moreover, as I previously said, you should rename the component with "popup" or "modal" keyword and rename this prop with a work like "isOpen"

src/app/js/admin/upload/ConfirmUpload.js
```diff
@@ -0,0 +1,85 @@
...
+    cancelUpload,
+    p: polyglot,
+    isOpenPopup,
+    onConfirm,
+}) => {
+    const handleConfirmAction = () => {
+        onConfirm();
+        cancelUpload();
+    };
+    const actions = [
```
jdemangeon: Why do you use a variable for actions instead of putting them directly inside the jsx tree?

src/app/js/admin/upload/ConfirmUpload.js
```diff
@@ -0,0 +1,85 @@
...
+        <Button
+            variant="text"
+            color="primary"
+            key="confirm"
+            className="confirm"
```
jdemangeon: Using classname for e2e tests is a bad practice, please use text inside it as selector instead

src/app/js/admin/upload/Upload.js
```diff
@@ -94,17 +103,30 @@ export const UploadComponent = ({
...
         if (files.length === 0) return;
+        if (!isFirstFile) {
+            onOpenUploadPopup();
+        } else {
+            onConfirm();
+        }
+    };
+
+    const onConfirm = (...params) => {
+        if (useUrl) {
```
jdemangeon: I don't see where the useUrl is declared :thinking: Moreover, this name seems to be too generic IMO

src/app/js/admin/upload/index.js
```diff
@@ -58,6 +63,14 @@ export default handleActions(
             ...state,
             open: false,
         }),
+        OPEN_UPLOAD_POPUP: state => ({
```
jdemangeon: Does this popup act as a global one? If not, you don't need to put it in redux. If yes, you should be sure that the component is "mounted" on the whole app container

jdemangeon (changes requested):

cypress/support/datasetImportPage.js
arimet: This is an arbitrary value used in other tests

src/app/js/admin/upload/ConfirmUpload.js
arimet: Rename in PopupConfirmUpload

src/app/js/admin/upload/ConfirmUpload.js
arimet: I applied the same architecture as the popup containing the duplicate URIs

src/app/js/admin/upload/index.js
arimet: I changed the logic to use state and props!

babel.config.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
                             browsers: ['last 2 versions'],
                         },
                         modules: false,
+                        corejs: '2.6.12',
```
jdemangeon: Hum... I'm not a big fan. Can you try again to remove it?

cypress/integration/publication.spec.js
```diff
@@ -31,6 +31,26 @@ describe('Dataset Publication', () => {
                 .eq(1)
                 .should('contains.text', ['2', 'Row 2', 'Test 2'].join(''));
         });
+
+        it('should display a information popup when adding a second dataset', () => {
+            menu.openAdvancedDrawer();
+            menu.goToAdminDashboard();
+            cy.wait(300);
+            datasetImportPage.importDataset('dataset/simple.csv');
+            cy.wait(500);
```
jdemangeon: Same remark about wait

src/app/js/admin/upload/PopupConfirmUpload.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+export const PopupConfirmUploadComponent = ({
+    p: polyglot,
+    isOpen,
+    onConfirm,
+    setIsOpenPopupConfirm
+}) => {
+
+    const closePopup = () => {
```
jdemangeon: Can you rename it to "handleClose"? (always prefix inner handler function with "handle")

https://jaketrent.com/post/naming-event-handlers-react
https://stackoverflow.com/questions/60048249/what-is-the-right-name-of-event-handler-onclick-or-handleclick

src/app/js/admin/upload/PopupConfirmUpload.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+PopupConfirmUploadComponent.propTypes = {
+    p: polyglotPropTypes.isRequired,
+    onConfirm: PropTypes.func.isRequired,
+    isOpen: PropTypes.bool.isRequired,
+    setIsOpenPopupConfirm: PropTypes.func.isRequired,
+};
+
+export default compose(
```
jdemangeon: You don't need compose here (because you only have one HOC)

src/app/js/admin/upload/Upload.js
```diff
@@ -144,7 +165,7 @@ export const UploadComponent = ({
                     }
                 />
             )}
-            {useUrl && (
+            {useUrlForUpload && (
```
jdemangeon, line 168: Can you put inner elements inside the same group? They have the same conditon (just put a fragment around them)

jdemangeon (changes requested):

jdemangeon (approved):

PR #1294, opened 2021-10-19
38: Fix Form Column Width

jdemangeon:
image


arimet (approved):

arimet:
What does it look like with a larger column name and data?

jdemangeon:
There's an ellipsis on datas ;)

PR #1295, opened 2021-10-19
41: Delete Field Creation From Subresource Column

jdemangeon:


src/app/js/admin/FieldsEdit.js
```diff
@@ -49,7 +49,7 @@ export const FieldsEditComponent = ({
             {tab === 'page' && (
                 <>
                     <div className={classes.actionsContainer}>
-                        {filter === SCOPE_DOCUMENT && (
+                        {filter === SCOPE_DOCUMENT && !subresourceId && (
```
arimet, line 52: I have trouble seeing the link between the development that is done and the associated ticket. 

If the error is coming from the `<AddFieldFromColumnButton>`, doesn't it also appear if you go through the selection in the popup?

jdemangeon: We can't select a "field" from a subresource because of the nature of subresources documents. So, I've removed the "show select column" button from the subresources page

arimet (approved):

PR #1296, opened 2021-10-20
6: Fix routine for subresource

arimet:


src/api/services/publishDocuments.spec.js
```diff
@@ -207,4 +208,138 @@ describe('publishDocuments', () => {
...
+                        },
+                    ],
+                },
+            ];
+
+            const result = getFieldsFromSubresourceFields(
+                fieldsFromSubressource,
+                true,
+            );
+            expect(result[0]).toHaveProperty('transformers');
```
jdemangeon: Useless, since you test on the line below

src/api/services/publishDocuments.spec.js
```diff
@@ -207,4 +208,138 @@ describe('publishDocuments', () => {
...
```
jdemangeon: Same

jdemangeon (changes requested):
Congrats, there's a lot of tests ;)

jdemangeon (approved):

PR #1312, opened 2021-11-02
Set a color to enriched columns

arimet:
image

### DIFF
--- a/src/app/js/admin/parsing/ParsingExcerpt.js
+++ b/src/app/js/admin/parsing/ParsingExcerpt.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import compose from 'recompose/compose';
@@ -23,6 +23,9 @@ const styles = {
     body: {
         position: 'relative',
     },
+    enrichedColumn: {
+        backgroundColor: theme.green.light,
+    },
 };
 
 export const getRowStyle = (index, total) => {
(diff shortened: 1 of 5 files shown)


src/app/js/admin/parsing/ParsingExcerpt.js
```diff
@@ -53,6 +69,11 @@ export const ParsingExcerptComponent = ({
                         <ParsingExcerptHeaderColumn
                             key={`header_${column}`}
                             column={column}
+                            style={
+                                arrayIsColumnEnrichment[column] && {
+                                    backgroundColor: theme.green.light,
```
guilbill: I think you'd better add an entry in the `styles` const, and reference it in here.

src/app/js/admin/parsing/ParsingExcerpt.js
```diff
@@ -69,10 +90,16 @@ export const ParsingExcerptComponent = ({
                                 showAddColumns &&
                                 (index === total - 3 ||
                                     (total < 3 && index === 0));
+
                             return (
                                 <ParsingExcerptColumn
                                     key={`${column}_${line._id}`}
                                     value={`${line[column] || ''}`}
+                                    style={
+                                        arrayIsColumnEnrichment[column] && {
```
guilbill: I think you'd better add an entry in the `styles` const, and reference it in here.

src/app/js/admin/parsing/ParsingExcerpt.js
```diff
@@ -37,13 +39,27 @@ export const getRowStyle = (index, total) => {
...
+
 export const ParsingExcerptComponent = ({
     columns,
     handleAddColumn,
     lines,
     showAddColumns,
+    enrichments,
 }) => {
     const total = lines.length;
+    const arrayIsColumnEnrichment = isColumnEnrichment(
```
guilbill: You could use `useMemo` here to avoid recomputing the array each time we render, but only when columns or enrichments changed. (https://fr.reactjs.org/docs/hooks-reference.html#usememo)

src/app/js/admin/parsing/ParsingExcerpt.js
```diff
@@ -37,13 +39,27 @@ export const getRowStyle = (index, total) => {
     return { opacity, height: 36 };
 };
 
+export const isColumnEnrichment = (columns, enrichments) => {
+    let arrayIsColumnEnrichment = [];
+    columns.map(c => {
```
guilbill: I prefer to have a descriptive variable, here, instead of `c`, it may be `columnName`? Same for enrichment, I know it's quite obvious that `e` is for `enrichiment` given that `enrichments` variable is not far, but IMO it's a good habit to have to name things with a clear, self explanatory name, even without the context.

src/app/js/admin/parsing/ParsingExcerpt.js
guilbill: That said, I'm not sure it's worth the (little) complexity it adds to create this `isColumnEnrichment`, it may be clearer, to "just" map the enrichment to have a list of names, and use the `includes` function to check if a column is an enrichment.

guilbill (approved):
Nice

PR #1316, opened 2021-11-05
Publication background job server

guilbill:


cypress/support/datasetImportPage.js
```diff
@@ -41,9 +41,7 @@ export const importMoreDataset = (filename, mimeType = 'text/csv') => {
     cy.get('#confirm-upload', { timeout: 1000 }).should('be.visible');
     cy.wait(300);
     cy.contains('Accept').click({ force: true });
-    cy.get('.progress').should('exist');
-    cy.get('.progress', { timeout: 6000 }).should('not.exist');
-
+    cy.wait(1000);
```
arimet: I don't know about you, but Julien told me that it was better to make a get with a timeout than to wait.

`cy.get('.data-published a', { timeout: 1000 }).contains`

src/api/controller/api/publish.js
```diff
@@ -1,64 +1,36 @@
...
-        await ctx.publishedDataset.remove({});
-        await ctx.publishedCharacteristic.remove({});
-        await ctx.publishedFacet.remove({});
-
+        await clearPublished(ctx);
         ctx.body = {
             status: 'success',
         };
     } catch (error) {
+        console.log(error);
```
arimet, line 25: forgotten or left on purpose?

src/app/js/admin/progress/reducer.js
```diff
@@ -36,13 +38,29 @@ export default handleActions(
...
         }),
-        [combineActions(PUBLISH, UPLOAD_FILE)]: state => ({
+        [UPLOAD_FILE]: state => ({
             ...state,
             status: STARTING,
             progress: undefined,
             target: undefined,
             error: undefined,
         }),
+        [PUBLISH]: state => ({
```
arimet, line 48: maybe [combineActions(PUBLISH, CLEAR_DATASET)] ?

src/app/js/lib/components/ButtonWithStatus.js
```diff
@@ -1,47 +1,93 @@
...
+                    variant="determinate"
+                    value={target ? (progress / target) * 100 : 0}
+                />
+            ) : (
+                undefined
+            )}
+        </div>
     );
+};
+const useStyles = makeStyles({
```
arimet: More a global question here, do we have conventions for the order of elements in a component? 
Because sometimes the style comes first, sometimes after the render etc...

arimet (approved):

cypress/support/datasetImportPage.js
floo51: Yes, because it returns earlier if the element is found and hence speeds up tests

src/api/controller/api/publish.js
guilbill: Left on purpose to trace the error

cypress/support/datasetImportPage.js
guilbill: godd advice, I change it

src/app/js/admin/progress/reducer.js
guilbill: It was a mistake, CLEAR_DATASET shouldn't be background

src/app/js/lib/components/ButtonWithStatus.js
guilbill: We should, what do you prefer? I'm not a very consistent guy... So I'll stick to what you prefer ;)

arimet (approved):

PR #1317, opened 2021-11-05
39: Add internal name for field + use this informations in fields selector.

arimet:
## Screenshot

image
image
image

### DIFF
--- a/src/app/js/fields/FieldInternalIcon.js
+++ b/src/app/js/fields/FieldInternalIcon.js
@@ -7,21 +7,21 @@ import BarChartIcon from '@material-ui/icons/BarChart';
 const HOME = 'home';
 const DOCUMENT = 'document';
 const CHART = 'chart';
-const INTERNAL_STATE_ICON = [];
-INTERNAL_STATE_ICON['home'] = <HomeIcon />;
-INTERNAL_STATE_ICON['document'] = <DescriptionIcon />;
-INTERNAL_STATE_ICON['chart'] = <BarChartIcon />;
+const INTERNAL_SCOPE_ICON = [];
+INTERNAL_SCOPE_ICON['home'] = <HomeIcon />;
+INTERNAL_SCOPE_ICON['document'] = <DescriptionIcon />;
+INTERNAL_SCOPE_ICON['chart'] = <BarChartIcon />;
 
-const getIconInternalState = state => {
-    if (!!state) {
-        return INTERNAL_STATE_ICON[state];
+const getIconInternalScope = scope => {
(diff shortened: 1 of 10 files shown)


src/app/js/fields/FieldInternalState.js
```diff
@@ -0,0 +1,73 @@
...
+const useStyles = makeStyles({
+    container: {
+        display: 'flex',
+        flexDirection: 'row',
+        alignItems: 'center',
+        marginRight: '1rem',
+    },
+    state: {
+        padding: '0.5rem',
+        border: '1px solid black',
```
guilbill: I've noticed that the black is not the right color, I think the color should be theme.black.secondary

guilbill (changes requested):
Well done for the "toggle" button !

Although I wonder why not having "Toggle" in the component name. One more thing to note is that I think it's a bad idea to have "state" in the component name also (state already have a meaning in React, etc.), maybe... Scope? (as it is how this info is named in the code as far as I understand...)

src/app/js/fields/FieldToggleInternalScope.js
```diff
@@ -0,0 +1,45 @@
...
+
+export const FieldToggleInternalScopeComponent = ({ input }) => {
+    const classes = useStyles();
+
+    const handleStateSelected = (event, state) => {
+        input.onChange(state === input.value ? '' : state);
+    };
+
+    return (
+        <ToggleButtonGroup
```
guilbill, line 25: Nice

src/app/js/fields/FieldInternalIcon.js
```diff
@@ -0,0 +1,27 @@
...
+INTERNAL_STATE_ICON['document'] = <DescriptionIcon />;
+INTERNAL_STATE_ICON['chart'] = <BarChartIcon />;
+
+const getIconInternalScope = state => {
+    if (!!state) {
+        return INTERNAL_STATE_ICON[state];
+    }
+};
+
+const FieldInternalIcon = ({ state }) => (getIconInternalScope(state));
```
guilbill: Is it still in use?

guilbill (changes requested):

src/app/js/fields/FieldInternalIcon.js
arimet: It is still in use,  it is the component which returns the good icon in the selection of the fields and in the datagrid. but the term state doesn't make sense anymore, I just changed it to scope

guilbill (approved):

PR #1322, opened 2021-11-10
Add simplified form for enrichment

arimet:
## Screensho
formSimplified
t


src/api/controller/api/enrichment.js
```diff
@@ -17,8 +18,8 @@ export const setup = async (ctx, next) => {
 };
 
 export const postEnrichment = async ctx => {
-    const newResource = ctx.request.body;
-    const result = await ctx.enrichment.create(newResource);
+    const newEnrichmentWithRule = await createEnrichmentRule(ctx);
```
guilbill: Shouldn't we have a try catch? `createEnrichmentRule` can raise an exception "Missing parameters"

src/api/services/enrichment/enrichment.js
```diff
@@ -0,0 +1,86 @@
...
+
+        let rule = getEnrichmentRuleModel(data, enrichment);
+
+        return {
+            name: enrichment.name,
+            advancedMode: enrichment.advancedMode,
+            rule: rule,
+        };
+    } catch (e) {
+        console.log('Error:', e.stack);
```
guilbill: I think there is something wrong here, if we catch an erro, then we return nothing, maybe we should just let the error propagate to the top level?

src/api/services/enrichment/enrichment.js
```diff
@@ -0,0 +1,86 @@
...
+                enrichment.webServiceUrl,
+            );
+        }
+
+        if (typeof sourceData === 'object') {
+            if (!enrichment.subPath) {
+                throw new Error(`Missing sub-path parameters`);
+            }
+            const subPathData = sourceData[0][enrichment.subPath];
+            if (!!!subPathData) {
```
guilbill: `!!!` ? Strange, I've never used this, is it a known pattern to check if the data is false?

guilbill (approved):

PR #1328, opened 2021-11-23
Update enrichment for direct path and multiple values

arimet:


src/api/workers/enrichment.js
```diff
@@ -60,11 +59,7 @@ const worker = async (id, ctx) => {
                 { $set: { [enrichment.name]: value } },
             );
         } catch (e) {
-            console.log({ e });
-            await ctx.dataset.updateOne(
-                { _id: new ObjectId(entry._id) },
-                { $set: { [enrichment.name]: '##ERROR##' } },
-            );
+            console.error({ e });
```
guilbill, line 62: Why removing the update ?

Edit : got it, to avoid using an undefined variable, but I think we should still set the error, without the enrichment.name.

src/api/services/enrichment/enrichment.js
```diff
@@ -32,13 +32,16 @@ export const createEnrichmentRule = async ctx => {
 export const getEnrichmentRuleModel = (sourceData, enrichment) => {
     try {
         let rule;
-        if (typeof sourceData === 'string' || (Array.isArray(sourceData) && typeof sourceData[0] === 'string')) {
+        if (
+            typeof sourceData === 'string' ||
```
guilbill: It could be clearer to extract those conditions in function, like :

```js
const isDirectPath = (sourceData) => {
    return typeof sourceData === 'string' ||
            (Array.isArray(sourceData) && typeof sourceData[0] === 'string');
}

```

src/api/services/enrichment/enrichment.js
```diff
@@ -47,7 +50,11 @@ export const getEnrichmentRuleModel = (sourceData, enrichment) => {
             );
         }
 
-        if (typeof sourceData === 'object' && (Array.isArray(sourceData)  && typeof sourceData[0] === 'object')) {
+        if (
+            typeof sourceData === 'object' &&
```
guilbill: Same here

guilbill (changes requested):
Good job! 
I think we should add some tests on enrichment.spec.js It's a critical part, and for now, there are tests only for successfull  case. We can improve on future dev, but do not forget to add them.

guilbill (approved):

PR #1338, opened 2021-12-01
Add tooltip for internal scope icon

arimet:


src/app/js/fields/FilterAt.js
```diff
@@ -1,16 +1,18 @@
 import * as React from 'react';
 import SvgIcon from '@material-ui/core/SvgIcon';
 
-export const FilterAt = props => {
+export const FilterAt = React.forwardRef((props, ref) => {
```
guilbill, line 4: What is the `ref` for ?

arimet: Mui Tooltip is not working properly with my custom icon `FilterAt`. It's because the child of a Material-UI Tooltip must be able to hold a ref. When i googled my probleme, I saw that using React.forwardRef would solve my problem

guilbill: Good to know! :)

guilbill (approved):

PR #1339, opened 2021-12-01
Check undedined return from WS

arimet:
## [Trello card](https://trello.com/c/xk8a3EYV/87-enrichissement-ini-vide)

- [x] Handles the misunderstood returns of the web service and displays them correctly. It also displays false values

### DIFF
--- a/src/api/workers/enrichment.js
+++ b/src/api/workers/enrichment.js
@@ -51,19 +51,27 @@ const worker = async (id, ctx) => {
 
     return async entry => {
         const enricherTransformer = await transformerFactory(id);
+        const enrichment = await ctx.enrichment.findOneById(id);
+        await ctx.dataset.updateOne(
+            { _id: new ObjectId(entry._id) },
+            { $set: { [enrichment.name]: 'En attente' } },
+        );
 
         try {
-            const { value, enrichment } = await enricherTransformer({
+            let { value } = await enricherTransformer({
                 id: entry._id,
                 value: entry,
             });
+
+            if (value === undefined) {
+                value = 'n/a';
+            }
 
             await ctx.dataset.updateOne(
                 { _id: new ObjectId(entry._id) },
                 { $set: { [enrichment.name]: value } },
             );
         } catch (e) {
-            const enrichment = await ctx.enrichment.findOneById(id);
             await ctx.dataset.updateOne(
                 { _id: new ObjectId(entry._id) },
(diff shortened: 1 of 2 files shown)


src/api/workers/enrichment.js
```diff
@@ -51,19 +51,27 @@ const worker = async (id, ctx) => {
 
     return async entry => {
         const enricherTransformer = await transformerFactory(id);
+        const enrichment = await ctx.enrichment.findOneById(id);
+        await ctx.dataset.updateOne(
+            { _id: new ObjectId(entry._id) },
+            { $set: { [enrichment.name]: 'En attente' } },
```
floo51, line 57: What about english translation for En attente?

src/api/workers/enrichment.js
```diff
@@ -51,19 +51,27 @@ const worker = async (id, ctx) => {
...
 
         try {
-            const { value, enrichment } = await enricherTransformer({
+            let { value } = await enricherTransformer({
                 id: entry._id,
                 value: entry,
             });
 
+            if (value === undefined) {
+                value = 'n/a';
```
floo51, line 67: Are we sure we want n/a in the field? If I remember correctly, front end doesn't display fields that have a null value.
Here it would show an error field instead.

src/app/js/admin/parsing/ParsingExcerptColumn.js
```diff
@@ -21,7 +21,7 @@ export const ParsingExcerptColumnComponent = ({ children, style, value }) =>
         </TableCell>
     ) : (
         <TableCell style={{ ...styles.col, ...style }}>
-            {value}
+            {`${value}`}
```
floo51, line 24: Is this a cast to string?

src/api/workers/enrichment.js
arimet: Yes, seen with the Lodex team on mattermost yesterday :)
For them if the value is null it means that nothing happened, so they didn't know if the webservice was called

src/app/js/admin/parsing/ParsingExcerptColumn.js
arimet: Yes, it allows us to display if a value is false

src/api/workers/enrichment.js
arimet: I didn't know how to manage the translation on the API side. Do I have to send the language used from the front end in the request and do the translation accordingly?

PR #1340, opened 2021-12-02
Fix the creation of an enrichment when the source is a number

arimet:
## [Trello card](https://trello.com/c/xk8a3EYV/87-enrichissement-ini-vide)

- [ ] We check that the line used is not empty. And we now take into account the number types

### DIFF
--- a/src/api/services/enrichment/enrichment.js
+++ b/src/api/services/enrichment/enrichment.js
@@ -12,7 +12,13 @@ export const createEnrichmentRule = async ctx => {
     }
 
     const excerpt = await ctx.dataset.getExcerpt();
-    const sourceData = excerpt[0][enrichment.sourceColumn];
+    let sourceData;
+    for (let line of excerpt) {
+        if (line[enrichment.sourceColumn]) {
+            sourceData = line[enrichment.sourceColumn];
+            break;
+        }
+    }
     let data;
     try {
         data = JSON.parse(sourceData);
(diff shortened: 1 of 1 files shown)


src/api/services/enrichment/enrichment.js
```diff
@@ -12,7 +12,13 @@ export const createEnrichmentRule = async ctx => {
     }
 
     const excerpt = await ctx.dataset.getExcerpt();
-    const sourceData = excerpt[0][enrichment.sourceColumn];
+    let sourceData;
+    for (let line of excerpt) {
```
floo51: What if the column is still empty in the whole excerpt?
I would add a specific method, something `findFirstByNotNull` in [the dataset model](https://github.com/Inist-CNRS/lodex/blob/0d26ca0d05f05b829edea3ca22b7cc0c79587766/src/api/models/dataset.js#L103)

floo51: You could use a `find`, something like:
```
const sourceData = lines.find((line)=> !!line[enrichment.sourceColumn]))
```

src/api/services/enrichment/enrichment.js
```diff
@@ -31,7 +37,10 @@ export const createEnrichmentRule = async ctx => {
 const isDirectPath = sourceData => {
     return (
         typeof sourceData === 'string' ||
-        (Array.isArray(sourceData) && typeof sourceData[0] === 'string')
+        typeof sourceData === 'number' ||
```
floo51: Can we make this more readable with functions and/or intermediate well named variables?

PR #1341, opened 2021-12-02
Add other selection for subressource column

arimet:
## [Trello card](https://trello.com/c/6r0588Ow/92-cr%C3%A9ation-nouveau-champ-liste-des-colonnes-pour-choisir-la-valeur-dun-champ)

- [x] Add Other field for subressource column selection. 

First iteration. I know that the hard value "Autre" is not great, but it is the only solution that came to mind in a simple way

## Screenshot

image

### DIFF
--- a/src/app/js/fields/wizard/SelectSubresourceField.js
+++ b/src/app/js/fields/wizard/SelectSubresourceField.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import {
     Select,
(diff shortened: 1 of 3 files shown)


src/app/js/fields/wizard/SelectSubresourceField.js
```diff
@@ -72,11 +106,14 @@ export const mapStateToProps = (state, { subresourceUri }) => {
...
-                ? subresourceData[0]
-                : subresourceData) || {},
-        ),
+        datasetFields: [
+            ...Object.keys(
+                (Array.isArray(subresourceData)
+                    ? subresourceData[0]
+                    : subresourceData) || {},
+            ),
+            ...['Autre'],
```
floo51: This needs to be translated

src/app/js/fields/wizard/SelectSubresourceField.js
```diff
@@ -22,28 +31,53 @@ export const SelectSubresourceFieldComponent = ({
...
-                <MenuItem
-                    key={`id_${datasetField}`}
-                    className={`column-${datasetField}`}
-                    value={datasetField}
+}) => {
+    const [isOtherColumn, setIsOtherColumn] = useState(false);
+
+    useEffect(() => {
+        if (column) {
+            if (datasetFields.includes(column) && column !== 'Autre') {
```
floo51: Put in a const if you use a magic string twice

src/app/js/fields/wizard/SelectSubresourceField.js
```diff
@@ -22,28 +31,53 @@ export const SelectSubresourceFieldComponent = ({
...
-        >
-            {datasetFields.map(datasetField => (
-                <MenuItem
-                    key={`id_${datasetField}`}
-                    className={`column-${datasetField}`}
-                    value={datasetField}
+}) => {
+    const [isOtherColumn, setIsOtherColumn] = useState(false);
+
+    useEffect(() => {
```
floo51: Is there a benefit of the `useEffect` over overriding given `handleChange`?

PR #1343, opened 2021-12-07
Increased user experience

arimet:
- [x] Remonter les boutons "nouveau champ" au même niveau que les tabs (supprimer les espaces rouges) 
 - [x] Nouveau champ depuis une colonne : mettre la datagrid dans une modal, même taille que la modale paramétrage d'un champ, titre de la modale : "Sélectionner une colonne"
  - [x] Fermer modale champ quand click sur backdrop
 - [x]  Faire que la zone de drag n drop des champs soit responsive (prenne 100% de l'espace disponible)
- [x] Quand on vient de créer un champ, scroll au niveau de ce champ
- [x] Sur l'import d'un modèle, ne pas mettre le warning si le modèle actuel est vide

  

- [x]  Enrichment :
    - mieux aligner les champs Source column et sub path
    - Aérer un peu le titre, la carte, et le bouton save
    - Revoir couleur "Mode avancé"


src/app/js/admin/Appbar/ClearDialog.js
```diff
@@ -38,6 +38,9 @@ const ClearDialogComponent = props => {
 
     useEffect(() => {
         if (succeeded) {
+            window.location.replace(
```
floo51: You sure we want that? It'll trigger a full page reload, not a react-router navigation

src/app/js/admin/Appbar/AddFieldFromColumnButton.js
```diff
@@ -45,7 +45,7 @@ export const AddFieldFromColumnButtonComponent = ({
 AddFieldFromColumnButtonComponent.propTypes = {
     onShowExistingColumns: PropTypes.func.isRequired,
     p: polyglotPropTypes.isRequired,
-    scope: PropTypes.string.isRequired,
```
floo51, line 48: Is this related to the change?

src/app/js/admin/ImportModelButton.js
```diff
@@ -22,31 +27,59 @@ const useStyles = makeStyles({
...
-    const [open, setOpen] = useState(false);
     const [
         showImportFieldsConfirmation,
         setShowImportFieldsConfirmation,
     ] = useState(false);
-    const [showSuccessAlert, setShowSuccessAlert] = useState(false);
 
-    const handleImportFieldsClose = sucess => {
+    const [showAlert, setShowAlert] = useState(false);
+    const [useInput, setUseInput] = useState(false);
```
floo51: Careful with this kind of naming in react, we could think it's a hook (and not the result of a hook)

src/app/js/admin/ImportModelButton.js
```diff
@@ -22,31 +27,59 @@ const useStyles = makeStyles({
...
     const classes = useStyles();
-    const [open, setOpen] = useState(false);
     const [
         showImportFieldsConfirmation,
         setShowImportFieldsConfirmation,
     ] = useState(false);
-    const [showSuccessAlert, setShowSuccessAlert] = useState(false);
 
-    const handleImportFieldsClose = sucess => {
+    const [showAlert, setShowAlert] = useState(false);
```
floo51: showAlert seems like a function, because it has a verb, but it's a status

src/app/js/admin/subresource/DeleteSubresourceButton.spec.js
```diff
@@ -57,7 +57,7 @@ describe('<DeleteSubresourceButton />', () => {
         );
 
         const okButton = wrapper.findWhere(node => {
-            return node.type() === Button && node.text() === 'OK';
+            return node.type() === Button && node.text() === 'Accept';
```
floo51, line 60: We sure about this? What about translations?

src/app/js/fields/FieldGrid.js
```diff
@@ -199,10 +207,30 @@ const DraggableItemGrid = ({
...
+
     useEffect(() => {
         setItems(buildFieldsDefinitionsArray(fields));
     }, [JSON.stringify(fields)]);
 
+    const fieldsLength = fields.length;
+    const previousFieldsLength = useRef(fields.length);
+    useEffect(() => {
+        if (previousFieldsLength.current + 1 === fields.length) {
+            setTimeout(() => {
```
floo51: This could be in a separate well-named function to make it clearer

src/app/js/fields/FieldGrid.js
```diff
@@ -88,6 +93,9 @@ const useStyles = makeStyles({
     },
 });
 
+// 1rem = 16px
+const ROOT_PADDING = 16;
```
floo51: Maybe we could take advantage of this to use the same variable in the styles.
Also, 1rem != 16px, it's relative to the font-size, font-size change = padding change

src/app/js/fields/wizard/SelectDatasetField.js
```diff
@@ -34,7 +34,7 @@ export const SelectDatasetFieldComponent = ({
             {datasetFields.map(datasetField => (
                 <MenuItem
                     key={`id_${datasetField}`}
-                    className={`column-${datasetField}`}
+                    className={`column-${datasetField.replaceAll(' ', '-')}`}
```
floo51, line 37: What if there are other characters like ' " \ ... ?

src/app/js/public/resource/Resource.js
```diff
@@ -85,7 +85,8 @@ export class ResourceComponent extends React.Component {
         if (
             match.params &&
             match.params.uri &&
-            match.params.uri.length !== 32 && // md5 length for subresources uuid
+            match.params.uri.length !== 32 &&
+            !match.params.uri.includes('%2F') && // md5 length for subresources uuid
```
floo51, line 89: Why not allowing a html encoded slash?

floo51 (changes requested):

src/app/js/admin/Appbar/AddFieldFromColumnButton.js
arimet: No, but it was an error that went up in the console. So I took the opportunity to correct it

src/app/js/admin/subresource/DeleteSubresourceButton.spec.js
arimet: So far I haven't seen any translations in the tests (maybe I missed them). It would be interesting to put them in place but not the subject of the ticket in my opinion

src/app/js/public/resource/Resource.js
arimet: html encoded slash is the old version of sub-resources. Because Julien made the change, we didn't think that it would affect the old models. So adding this condition allows us to check both possible cases.

PR #1344, opened 2021-12-07
Remove progress loader when error on clear data

arimet:


floo51 (commented):
It's missing something or the actions are already dispatched somewhere else?

arimet:
> It's missing something or the actions are already dispatched somewhere else?

Already dispatched somewhere else, Here it is just to reset the progress in case of error.

PR #1354, opened 2021-12-21
Add logs view for enrichment.

arimet:
image


src/app/js/admin/enrichment/enrichment-context.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
guilbill: You should rename the file, to match the camelCase convention

src/app/js/admin/enrichment/EnrichmentLogs.js
```diff
@@ -0,0 +1,117 @@
...
+        method: 'GET',
+    });
+
+    const [error, setError] = useState(null);
+    const [isLoaded, setIsLoaded] = useState(false);
+    const [logs, setLogs] = useState([]);
+
+    useEffect(() => {
+        enrichment?.jobId &&
+            fetch(url, options)
```
guilbill, line 56: I would extract this fetch in /js/lib as we will reuse this. Maybe a method like : `fetchData(url, method, payload)`

guilbill (approved):

PR #1360, opened 2021-12-23
Cancel publishing

guilbill:
## What is it ?

Add a cancel button to abort a running publication.

https://user-images.githubusercontent.com/560852/148070765-93720190-294c-4d02-b356-7ab1150f318f.mp4


src/api/controller/api/job.js
```diff
@@ -1,7 +1,10 @@
 import Koa from 'koa';
 import route from 'koa-route';
 import koaBodyParser from 'koa-bodyparser';
-import { enrichmentQueue } from '../../workers/enrichment';
+import { enrichmentQueue, ENRICHMENT_QUEUE } from '../../workers/enrichment';
```
floo51: Maybe not related to this PR but there's a naming issue. It's either publisher/enricher or publishment/enrichment

src/api/controller/api/job.js
```diff
@@ -13,14 +16,29 @@ export const setup = async (ctx, next) => {
...
 export const getJobLogs = async (ctx, queue, id) => {
-    if (queue === 'enrichment') {
+    if (queue === ENRICHMENT_QUEUE) {
         ctx.body = await enrichmentQueue.getJobLogs(id);
     }
 };
 
+export const cancelJob = async (ctx, queue) => {
+    if (queue === PUBLISHER_QUEUE) {
+        await publisherQueue.clean(0, 'wait');
```
floo51: What's 0? Can we make it clear by naming that arg?

src/api/controller/api/job.js
```diff
@@ -13,14 +16,29 @@ export const setup = async (ctx, next) => {
...
+        await publisherQueue.getActive().then(jobs => {
```
floo51: await and then in the same place?

src/api/services/indexSearchableFields.js
```diff
@@ -2,8 +2,11 @@ import composeAsync from '../../common/lib/composeAsync';
 import mongoClient from './mongoClient';
 import field from '../models/field';
 import publishedDataset from '../models/publishedDataset';
+import progress from './progress';
+import { CREATE_INDEX } from '../../common/progressStatus';
 
 export default async () => {
+    progress.start(CREATE_INDEX, null, null, 'publishing', null, 'publisher');
```
floo51: Very hard to read as we have no idea what args progress.start takes

src/api/services/publish.js
```diff
@@ -9,6 +8,8 @@ import {
     SCOPE_GRAPHIC,
 } from '../../common/scope';
 
+const isJobActive = async ctx => !ctx.job || (await ctx.job.isActive());
```
floo51, line 11: This function either returns a boolean or a promise, isn't it a code smell to have 2 different return types?

src/api/services/publish.js
```diff
@@ -17,19 +18,19 @@ export default async ctx => {
         [SCOPE_COLLECTION, SCOPE_DOCUMENT].includes(c.scope),
     );
 
-    await clearPublished(ctx);
-    await ctx.publishDocuments(ctx, count, collectionScopeFields);
+    (await isJobActive(ctx)) && (await clearPublished(ctx));
```
floo51: Why is isJobActive called for every line? Can we make it a big if on top? Or is it meant to be able to stop at any step of the publication?
Also, there's no return condition, which means if it's not active it'll try all the subsequent calls? and finish anyway?

floo51: If isJobActive should legit be called for every operation, then make it a bit more readable with a function like "chainOperationsWhileJobActive" ?

src/api/services/publishDocuments.js
```diff
@@ -203,8 +204,13 @@ export const publishDocumentsFactory = ({
         ctx.dataset.findLimitFromSkip,
         ctx.publishedDataset.insertBatch,
         versionTransformerDecorator(transformMainResourceDocument),
+        undefined,
+        ctx.job,
     );
-    ctx.job && ctx.job.log('Document published');
+    ctx.job &&
```
floo51: Is there an actual case when there's no ctx.job? Shouldn't it early return at the start of the method if so?

src/api/services/transformAllDocuments.js
```diff
@@ -6,9 +6,10 @@ export default async function transformAllDocument(
     insertBatch,
     transformer,
     datasetChunkExtractor = x => x,
+    job,
 ) {
     let handled = 0;
-    while (handled < count) {
+    while (handled < count && (!job || (job && (await job.isActive())))) {
```
floo51: This would be a lot clearer with a break condition at the end of the loop imo

src/app/js/admin/Appbar/JobProgress.js
```diff
@@ -109,6 +117,18 @@ const JobProgressComponent = props => {
                             </Typography>
                         )}
                     </div>
+
+                    {progress?.type === 'publisher' && (
+                        <Button
```
floo51: I don't see a pointer cursor on the video, is it only the video or does this button miss some styles?

src/app/js/admin/api/job.js
```diff
@@ -0,0 +1,21 @@
+import fetch from '../../lib/fetch';
+import { getJobLogsRequest, getCancelJobRequest } from '../../user';
+
+export const getUserLocalStorageInfo = () =>
+    JSON.parse(window.localStorage.getItem('redux-localstorage').user);
+
+const getJobLogs = (queue, jobId) => {
+    const state = getUserLocalStorageInfo();
+
+    const request = getJobLogsRequest(state, queue, jobId);
```
floo51: What's in the state? Why do we send the full state? Can't we cherry-pick only what we need from the state?

floo51 (changes requested):

floo51:
What happens when we cancel publication if it's already published? Does the front empty? There should be a confirm message if so. See mockup here: https://trello.com/c/5ulW993X/47-mise-en-place-dun-background-job-server-pour-la-publication

src/api/services/publish.js
guilbill: I'm currently testing this, and it always return a Promise
```
function resolveAfter2Seconds() {
   return new Promise(resolve => {
     setTimeout(() => {
       resolve('resolved');
     }, 2000);
   }); }

> const testBoolOrAwait = async (isDone) => isDone || await resolveAfter2Seconds();
undefined
> testBoolOrAwait()
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 4449,
  [Symbol(trigger_async_id_symbol)]: 5,
  [Symbol(destroyed)]: { destroyed: false }
}
> testBoolOrAwait(true)
Promise {
  true,
  [Symbol(async_id_symbol)]: 4483,
  [Symbol(trigger_async_id_symbol)]: 5,
  [Symbol(destroyed)]: { destroyed: false }
}
> testBoolOrAwait().then(d=>console.log(d))
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 4604,
  [Symbol(trigger_async_id_symbol)]: 4600,
  [Symbol(destroyed)]: { destroyed: false }
}
> resolved
> testBoolOrAwait(true).then(d=>console.log(d))
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 4709,
  [Symbol(trigger_async_id_symbol)]: 4708,
  [Symbol(destroyed)]: { destroyed: false }
}
> true
```

src/api/services/publishDocuments.js
```diff
@@ -85,7 +85,10 @@ export const publishDocumentsFactory = ({
     getDocumentTransformer,
     transformAllDocuments,
 }) => async (ctx, count, fields) => {
-    ctx.job && ctx.job.log('Publishing documents');
+    if (!ctx.job) {
```
floo51, line 90: We might want to log somewhere if this happens?

src/api/services/publish.js
```diff
@@ -18,19 +29,26 @@ export default async ctx => {
         [SCOPE_COLLECTION, SCOPE_DOCUMENT].includes(c.scope),
     );
 
-    (await isJobActive(ctx)) && (await clearPublished(ctx));
-    (await isJobActive(ctx)) &&
-        (await ctx.publishDocuments(ctx, count, collectionScopeFields));
-
-    const datasetScopeFields = fields.filter(c =>
-        [SCOPE_DATASET, SCOPE_GRAPHIC].includes(c.scope),
+    await chainWhileJobIsActive(
```
floo51, line 32: 😍

src/api/services/transformAllDocuments.js
```diff
@@ -9,7 +9,11 @@ export default async function transformAllDocument(
     job,
 ) {
     let handled = 0;
-    while (handled < count && (!job || (job && (await job.isActive())))) {
+    const isJobActive = async () =>
+        !job || (job.isActive && (await job.isActive()));
+    while (handled < count) {
+        if (!(await isJobActive())) break;
```
floo51: Did we eslint this? Extra parentheses and no brackets looks weird to me?

src/app/js/admin/api/job.js
```diff
@@ -5,9 +5,9 @@ export const getUserLocalStorageInfo = () =>
     JSON.parse(window.localStorage.getItem('redux-localstorage').user);
 
 const getJobLogs = (queue, jobId) => {
-    const state = getUserLocalStorageInfo();
+    const userInfo = getUserLocalStorageInfo();
 
-    const request = getJobLogsRequest(state, queue, jobId);
+    const request = getJobLogsRequest(userInfo, queue, jobId);
```
floo51: It's not clearer to me what's in this object and why we send it all

floo51: Looking at the function quickly, it seems we only need the token, why not send this only?

floo51 (changes requested):

src/api/services/transformAllDocuments.js
guilbill: This is eslint(with prettier) that added the extra parentheses, I add brackets for clarity. I find it weird to...

src/api/services/publishDocuments.js
guilbill: At first I was thinking that we could just log in the job (bull allow to add logs in jobs), but I can add logs in our logger, so that we see what happen without looking at the job logs.

floo51: Maybe a detailed technical log in bull and a more generic "Didnt work, see devs" in lodex?

floo51 (approved):

PR #1366, opened 2022-01-06
Fix dataset display

guilbill:
https://user-images.githubusercontent.com/560852/148409875-d8a8c83a-ce0c-4f54-94e1-ce6676756eab.mp4


floo51 (approved):

floo51:
How comes some number appear with quotes on your video? Guess it's because they're actual strings in the dataset, but then again is it an import problem or a dataset problem?

guilbill:
After checking, the data is like that from the beginning.

PR #1375, opened 2022-01-20
Sequential jobs

guilbill:
Jobs are now running sequentialy.

https://user-images.githubusercontent.com/560852/150301248-fc4555a2-3d6e-4d37-9f34-89169af3b448.mp4

### DIFF
--- a/src/api/controller/api/job.js
+++ b/src/api/controller/api/job.js
@@ -1,8 +1,7 @@
 import Koa from 'koa';
 import route from 'koa-route';
 import koaBodyParser from 'koa-bodyparser';
-import { enricherQueue, ENRICHER_QUEUE } from '../../workers/enricher';
-import { publisherQueue, PUBLISHER_QUEUE } from '../../workers/publisher';
+import { workerQueue, QUEUE_NAME } from '../../workers/publisher';
 import clearPublished from '../../services/clearPublished';
 import progress from '../../services/progress';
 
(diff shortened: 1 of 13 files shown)


src/api/workers/publisher.js
```diff
@@ -1,27 +1,15 @@
...
-            removeOnFail: 100,
-            lifo: true,
-        },
-    },
-);
 
-publisherQueue.process(PUBLISH, (job, done) => {
-    publisherQueue.clean(100, 'wait');
+export const processPublication = (job, done) => {
+    cleanWaitingJobsOfType(PUBLISHER);
```
floo51, line 12: No await here?

src/api/controller/api/job.js
```diff
@@ -16,30 +14,26 @@ export const setup = async (ctx, next) => {
...
-        await publisherQueue.clean(GRACE_PERIOD, 'wait');
-        const activeJobs = await publisherQueue.getActive();
-        activeJobs.forEach(job => {
-            job.moveToFailed(new Error('cancelled'), true);
-        });
-
+export const cancelJob = async (ctx, type) => {
+    const activeJob = (await workerQueue.getActive())[0];
+    if (activeJob?.data?.jobType === type) {
+        await cleanWaitingJobsOfType(activeJob.data.jobType);
```
floo51, line 24: Why await here?

floo51 (changes requested):

src/api/controller/api/job.js
guilbill: Here I need to be sure that my waiting jobs are cleaned before moving my active job to 'Failed', otherwise one of the waiting job may be placed immediately as active.

src/api/workers/publisher.js
guilbill: Here, I don't need to wait, the active job is just starting, so there is no risk than one waiting job, go active before beeing cleaned.

floo51 (approved):

PR #1385, opened 2022-01-28
Delete enrichment

guilbill:
https://user-images.githubusercontent.com/560852/151489910-cd3042fc-e62a-42d1-89a9-eb44159f6e72.mp4

### DIFF
--- a/src/api/controller/api/enrichment.js
+++ b/src/api/controller/api/enrichment.js
@@ -9,8 +9,7 @@ import {
     getEnrichmentDataPreview,
     setEnrichmentJobId,
 } from '../../services/enrichment/enrichment';
-import { cancelJob } from './job';
-import { getActiveJob } from '../../workers/tools';
+import { cancelJob, getActiveJob } from '../../workers/tools';
 
 export const setup = async (ctx, next) => {
     try {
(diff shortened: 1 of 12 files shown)


src/api/controller/api/enrichment.js
```diff
@@ -57,6 +59,13 @@ export const putEnrichment = async (ctx, id) => {
 export const deleteEnrichment = async (ctx, id) => {
     try {
         const enrichment = await ctx.enrichment.findOneById(id);
+        const activeJob = await getActiveJob();
+        if (
+            activeJob?.data?.jobType === ENRICHER &&
+            activeJob?.data?.id === id
+        ) {
+            cancelJob(ctx, ENRICHER);
```
floo51, line 68: This is called asynchronously and modified `ctx.status`, possibly after we already answered. I have the feeling this could break. Can we have a cancelJob function that does just that and let the caller handle return codes?

floo51 (changes requested):

src/api/controller/api/enrichment.js
guilbill: Yes, you're right, well spotted

floo51 (approved):

PR #1388, opened 2022-01-31
use latest ezs packages

touv:


touv:
see unit test here https://github.com/Inist-CNRS/ezs/pull/209

touv:
tmp file was deleted when the stream was closed. With no back pressure control, we can write in closed stream, and so we can try to access to non existing file. With this patch there is an error before it occurs

guilbill (approved):

PR #1429, opened 2022-04-14
Set a new version for the loaders

parmentf:
In order to access to the last json-conditor loader.


parmentf:
I don't understand why some tests are broken (but it seems that other pull requests have the same problem).
These modifications should not break anything.

touv:
see PR #1433

PR #1446, opened 2022-09-20
upd ezs/*

touv:


package.json
```diff
@@ -60,18 +60,18 @@
...
-        "@ezs/sparql": "^1.2.0",
-        "@ezs/storage": "^1.6.2",
+        "@ezs/conditor": "^2.8.2",
+        "@ezs/core": "2.0.1",
+        "@ezs/istex": "1.5.3",
+        "@ezs/lodex": "1.15.2",
+        "@ezs/loterre": "^2.0.0",
+        "@ezs/sparql": "^1.2.1",
+        "@ezs/storage": "^1.7.0",
         "@ezs/transformers": "1.3.11",
```
parmentf, line 73:
```suggestion
        "@ezs/transformers": "1.4.1",
```

Why not update to the last version of `@ezs/transformers` ?

parmentf (changes requested):

package.json
touv: Because of the change related to URI generation. A patch is need to update...

PR #1479, opened 2022-12-12
Feat: Add save as button for edit Cell

arimet:
https://user-images.githubusercontent.com/18294269/207098179-1b09cb19-67ff-447b-9928-ae5cfcec9fce.mp4

### DIFF
--- a/src/app/js/admin/parsing/ParsingEditCell.js
+++ b/src/app/js/admin/parsing/ParsingEditCell.js
@@ -15,10 +15,9 @@ import {
     Input,
     Menu,
     MenuItem,
-    Snackbar,
 } from '@material-ui/core';
-import { Alert } from '@material-ui/lab';
 import theme from '../../theme';
+import { toast } from 'react-toastify';
 
 const style = {
     container: {
@@ -81,6 +80,10 @@ export const getValueBySavingType = (value, type, previousValue) => {
         return Number(value);
     }
     if (type === 'string') {
+        const parsedValue = tryParseJSONObjectOrArray(value);
+        if (parsedValue && typeof parsedValue === 'object') {
+            return JSON.stringify(value);
+        }
         return String(value);
     }
     if (type === 'boolean') {
@@ -92,7 +95,7 @@ export const getValueBySavingType = (value, type, previousValue) => {
         if (!parsedValue || !(parsedValue instanceof Array)) {
             throw new Error('value_not_an_array');
         }
-        return value;
+        return parsedValue;
     }
 
     if (type === 'json') {
@@ -100,7 +103,7 @@ export const getValueBySavingType = (value, type, previousValue) => {
         if (!parsedValue || typeof parsedValue !== 'object') {
             throw new Error('value_not_an_object');
         }
-        return value;
+        return parsedValue;
     }
 
     // If no type is provided, we try to guess the type
@@ -199,15 +202,6 @@ ButtonWithDropdown.propTypes = {
 const ParsingEditCell = ({ cell, p: polyglot, setToggleDrawer }) => {
     const [loading, setLoading] = React.useState(false);
     const [value, setValue] = React.useState(returnParsedValue(cell.value));
-    const [openAlert, setOpenAlert] = React.useState(false);
-
-    const handleCloseAlert = (event, reason) => {
-        if (reason === 'clickaway') {
-            return;
-        }
-
-        setOpenAlert(false);
-    };
 
     const classes = useStyles();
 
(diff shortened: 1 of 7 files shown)


src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
     return value !== Object(value);
 };
 
+export const tryParseJSONObjectOrArray = jsonString => {
+    try {
+        if (typeof jsonString === 'object') {
+            return jsonString;
+        }
+        const parsed = JSON.parse(jsonString);
+        if (parsed && (typeof parsed === 'object' || parsed instanceof Array)) {
```
thibault-barrat: typeof an array will be object, so the condition `parsed instanceof Array` will never be used

src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
 };
 
-const isPrimitive = value => {
+export const isPrimitive = value => {
     return value !== Object(value);
 };
 
+export const tryParseJSONObjectOrArray = jsonString => {
+    try {
+        if (typeof jsonString === 'object') {
```
thibault-barrat, line 60: is jsonString can be null ? because typeof something null will be an object

src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
+    return false;
+};
+
+export const getValueBySavingType = (value, type, previousValue) => {
+    if (type === 'number') {
+        const parsedValue = Number(value);
+        if (isNaN(parsedValue)) {
+            throw new Error('value_not_a_number');
+        }
+        return Number(value);
```
thibault-barrat:
```suggestion
        return parsedValue;
```

src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
+    }
+    if (type === 'string') {
+        const parsedValue = tryParseJSONObjectOrArray(value);
+        if (parsedValue && typeof parsedValue === 'object') {
```
thibault-barrat: I think you already checked that parsedValue is an object in `tryParseJSONObjectOrArray`

src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
+            return JSON.stringify(value);
+        }
+        return String(value);
+    }
+    if (type === 'boolean') {
+        return value === 'true';
```
thibault-barrat: what if we want to convert have number 1 equals to true ? In lodex doc about BOOLEAN transformer (https://lodex.inist.fr/docs/partie-2-2/transformer-des-donnees/#BOOLEAN) : 
Toute chaîne de caractères valant true, 1, on, ok, oui, yes ou true sera vraie, et toute autre chaîne sera fausse.

src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
+        if (!parsedValue || typeof parsedValue !== 'object') {
+            throw new Error('value_not_an_object');
+        }
+        return parsedValue;
+    }
+
+    // If no type is provided, we try to guess the type
+    if (value instanceof Array || value instanceof Object) {
+        console.log('typeof previousValue', typeof previousValue);
+        console.log('previousValue', previousValue);
```
thibault-barrat: console.log spotted :)

src/app/js/admin/parsing/ParsingEditCell.js
```diff
@@ -28,71 +30,207 @@ const style = {
...
+        if (typeof previousValue === 'number') {
+            return Number(value);
+        }
+        if (typeof previousValue === 'boolean') {
+            return value === 'true';
+        }
+    }
+
+    return JSON.stringify(value);
+};
```
thibault-barrat:
```suggestion
    if ((value instanceof Array || value instanceof Object) && typeof previousValue !== 'string') {
        return value;
    } else if (isPrimitive(previousValue)) {
        if (typeof previousValue === 'number') {
            return Number(value);
        }
        if (typeof previousValue === 'boolean') {
            return value === 'true';
        }
    }

    return JSON.stringify(value);
};
```

thibault-barrat (changes requested):

src/app/js/admin/parsing/ParsingEditCell.js
arimet: If jsonString is null, it will return himself.

thibault-barrat (approved):

PR #1487, opened 2022-12-19
Feat: Add duplicate field

arimet:


cypress/integration/publication.spec.js
```diff
@@ -344,15 +344,15 @@ describe('Dataset Publication', () => {
                 .should('have.length', 3);
         });
     });
-    describe('Automatic Publication', () => {
+    describe.only('Automatic Publication', () => {
```
thibault-barrat: I don't know if `only` is working but you should remove it

thibault-barrat (changes requested):

thibault-barrat (approved):

PR #1522, opened 2023-01-09
Feat: replace alert by toast

arimet:


src/app/js/admin/Appbar/JobProgress.js
```diff
@@ -107,7 +107,7 @@ const JobProgressComponent = props => {
             if (data.success) {
                 handlePublishSuccess();
                 setProgress();
-            } else {
+            } else if (!data.isPublishing && !data.success) {
```
thibault-barrat: As you are in else, `!data.success` condition is useless here

thibault-barrat (approved):

PR #1553, opened 2023-01-24
Fix issues

arimet:


src/app/js/fields/FieldGrid.js
```diff
@@ -59,14 +63,16 @@ const useStyles = makeStyles({
         '&:hover': {
             background: '#efefef',
             // display the duplicate icon
```
thibault-barrat: display duplicate and setting icon

thibault-barrat (approved):

PR #1579, opened 2023-02-03
Feat: not unpublish before first publish

thibault-barrat:


src/app/js/admin/publish/sagas.js
```diff
@@ -15,18 +14,6 @@ export function* handlePublishRequest() {
             yield put(publishError(error));
             return;
         }
-
-        const { progressError } = yield race({
```
arimet, line 19: Nice but you didn't update the test since you remove this part.

arimet (approved):

PR #1594, opened 2023-02-10
Feat: disable auto close on error toast

thibault-barrat:
https://user-images.githubusercontent.com/41492953/218154638-ff26c14c-d4fb-4490-8df6-8e0b02da1af9.mov


arimet (approved):
I approve, but the demand is only for error ? 
Because the code could be simpler if we do the same for all notification

thibault-barrat:
> I approve, but the demand is only for error ? Because the code could be simpler if we do the same for all notification

Yes, the demand is only for error. That's why I modified each toast and not the container

floo51:
Can we proxy react-toastify with a function that calls it with the right arguments depending on error type? Code would be a lot cleaner

thibault-barrat:
> Can we proxy react-toastify with a function that calls it with the right arguments depending on error type? Code would be a lot cleaner

Smart idea 👍

floo51 (approved):

PR #1593, opened 2023-02-10
Feat: standardize action buttons

thibault-barrat:
image

PR #1595, opened 2023-02-10
Feat: add download enrichment logs button

thibault-barrat:
https://user-images.githubusercontent.com/41492953/218156343-9ec879bc-746e-4041-9729-acba72edecdb.mov


src/app/js/admin/enrichment/EnrichmentLogsDialog.js
```diff
@@ -87,6 +87,15 @@ export const EnrichmentLogsDialog = ({
 }) => {
     const [logsContainerRef, { width }] = useMeasure();
 
+    const handleDownloadLogs = () => {
+        const file = new Blob(logs, { type: 'text/plain' });
+        const element = document.createElement('a');
+        element.href = URL.createObjectURL(file);
+        element.download = 'enrichment-logs-' + Date.now() + '.log';
+        document.body.appendChild(element);
+        element.click();
```
arimet, line 97: I'm not sure that this solution work on chromium, could you try it ?

arimet (commented):
Conflict due to the merge of #1593

src/app/js/admin/enrichment/EnrichmentLogsDialog.js
thibault-barrat: Successfully tested on Chrome 👍

floo51:
Can we proxy react-toastify with a function that calls it with the right arguments depending on error type? Code would be a lot cleaner

src/app/js/admin/enrichment/EnrichmentLogsDialog.js
floo51: Is there a reason not to use a link with href styled as a button instead of a button that creates a hidden link with href?

floo51: You probably want to make it invisible, or to remove it from the DOM once clicked

thibault-barrat: I think it is better to create the file on click because you can open the log dialog at the beginning of the enrichment process. If you create the file at the render, the logs could be not complete.

floo51: Can we still remove `element` from `body` once we're done though?

floo51 (approved):

PR #1597, opened 2023-02-13
Feat: republish after deleting a dataset row

thibault-barrat:
https://user-images.githubusercontent.com/41492953/218489693-4caec1eb-6fb9-4503-801c-6da1d1b097bd.mov


arimet (approved):
sorry, I forgot this part because I had not published any data

PR #1598, opened 2023-02-14
Feat: add FieldAddDropdownButton for subresources

thibault-barrat:
https://user-images.githubusercontent.com/41492953/218785959-00727b1b-b55c-4549-a020-19c118562c8b.mov


src/app/js/admin/field/AddFromColumnDialog.js
```diff
@@ -7,34 +7,24 @@ import {
...
-                <div style={styles.container}>
-                    <ParsingResult
-                        showAddFromColumn
-                        maxLines={6}
-                        onAddField={onClose}
-                    />
-                </div>
+            <DialogContent
+                sx={{
+                    height: 'calc(90vh - 64px - 52px)',
```
floo51: We could replace those magic numbers by named variables, so that we know what 64px is the height of, in case we update it. Bonus if it is imported from the component whose height is 64px

src/app/js/fields/index.js
```diff
@@ -138,13 +164,18 @@ const getDefaultField = (name, index, rest = {}) => ({
 export default handleActions(
     {
         ADD_FIELD: (state, { payload }) => {
-            const { name, ...rest } = payload || {};
+            const { name, subresource, ...rest } = payload || {};
```
floo51: Avoid passing ...rest like arguments through, instead pick the variables you need from it.
Passing it fully makes it less maintainable and robust, if the payload changes because of some side effect, like adding an _id property, it would probably break mongoose.

src/app/js/admin/field/FieldAddDropdownButton.js
```diff
@@ -21,27 +21,29 @@ import { fromFields } from '../../sharedSelectors';
...
 
-const FieldAddDropdownButton = ({
+export const FieldAddDropdownButtonComponent = ({
     onAddNewField,
     onShowExistingColumns,
     isFieldsLoading,
     p: polyglot,
 }) => {
     const [open, setOpen] = React.useState(false);
+    const { subresourceId } = useParams();
```
floo51: This component could remain a dumb component if `subresourceId` was handled by one of the parent components.

src/app/js/admin/parsing/ParsingExcerptAddColumn.js
```diff
@@ -37,8 +24,16 @@ export const ParsingExcerptAddColumnComponent = ({
         )}`}
         onClick={handleAddColumn}
         color="primary"
-        style={styles.button(atTop)}
         disabled={isFieldsLoading}
+        sx={{
+            bottom: atTop ? '0' : '-68px',
```
floo51: Magic number here too

floo51 (approved):

floo51 (approved):

PR #1613, opened 2023-02-27
Feat: add dependency review and codeQL workflows

thibault-barrat:


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,76 @@
+# For most projects, this workflow file will not need changing; you simply need
```
arimet: Maybe this comment is not usefull

arimet (approved):

PR #1630, opened 2023-03-09
Feat: Add pdf to export dataset

arimet:


src/api/controller/api/pdf.js
```diff
@@ -0,0 +1,158 @@
...
+    5: 'Juin',
+    6: 'Juillet',
+    7: 'Août',
+    8: 'Septembre',
+    9: 'Octobre',
+    10: 'Novembre',
+    11: 'Décembre',
+};
+
+function getDateInFrench(date) {
```
thibault-barrat: What if the user use english ? Maybe we can get the locale used in front and use moment library to format the date accordingly ?

src/api/controller/api/pdf.js
```diff
@@ -0,0 +1,158 @@
...
+
+        // Create a document
+        const doc = new PDFDocument();
+
+        // Pipe its output somewhere, like to a file or HTTP response
+        doc.pipe(fs.createWriteStream('/tmp/publication.pdf'));
+
+        doc.font('Helvetica-Bold')
+            .fontSize(25)
+            .text('Données publiées', { align: 'center' });
```
thibault-barrat: Smae than date, if we can translate it according to the locale it should be great

thibault-barrat (changes requested):

thibault-barrat (approved):

touv:
How it happens if there are several thousand documents ? Is there a limit? Is there a way to disable/enable this new option ?

arimet:
We have added a limit that can be modified via the config. 
The display of the option is also done through the config. 

We can imagine in the future an interface to manage this point

thibault-barrat (approved):

PR #1666, opened 2023-03-24
Feat:fix encoded characters

touv:


touv:
see https://github.com/Inist-CNRS/lodex/pull/1667

PR #1693, opened 2023-08-16
skip parameter is usefull for kbart scroll

touv:


touv:
I don't know why skip parameter was dropped in a previous commit. It seem to fix a bug with Mongo 4. This patch restore the parameter, perhaps a bug will be back... or not

PR #1695, opened 2023-08-21
import themes in the same repo

touv:
Until now Lodex themes were stored in a separate repository (lodex-themes).
The idea would be to put the themes in the same repository and so  modifications and developments will be  always synchronized with the latest version.


parmentf:
It seems to be a mere storage?
You did not add any code to move the theme in the `custom` directory, for example (in a github action which could build a thematized docker image).

PR #1701, opened 2023-09-07
feat/admin: add external resource field

JulienMattiussi:
- Include transformers package in lodex
- Add WEBSERVICE transformer
- Add Webservice button in model interface

Next PR : Add webservice calls and treatment in publication step


packages/transformers/README.md
arimet, line 1: I was wondering if we should move these packages. Do the config files still make sense here? 

Also, should the changelog still exist. Shouldn't it then cohabit with the Lodex changelog (on Github)? 

Because transformers will now have to be released via Lodex and no longer from their separate packages, won't they? 

Then I think the Readme will have to be rewritten too.

Maybe ask Nicolas.

packages/transformers/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "license": "MIT",
+    "main": "./src/index.js",
+    "peerDependencies": {
+        "@ezs/core": "*"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "repository": "Inist-CNRS/ezs.git",
+    "scripts": {
```
arimet, line 30: What happens to this file? Isn't it up to Lodex to launch these orders afterwards?

JulienMattiussi: The install is managed by postinstall in the main package.json
But I think these local scripts can be usefull too for dev or test

arimet (approved):

PR #1715, opened 2023-09-27
Update to node 16

AlasDiablo:


parmentf:
```log
#11 98.76 > transformers@1.4.4 build
#11 98.76 > babel --root-mode upward src --out-dir lib
#11 98.76 
#11 99.04 Error: Babel was run with rootMode:"upward" but a root could not be found when searching upward from "/app/packages/transformers".
#11 99.04 One of the following config files must be in the directory tree: "babel.config.js, babel.config.cjs, babel.config.mjs, babel.config.json".
```

Possible that the error comes from the fact the `@ezs/transformers` was in a monorepo, where the [`babel.config.js`](https://github.com/Inist-CNRS/ezs/blob/master/babel.config.js) is at the root of the repo.

touv (commented):
use strict version

AlasDiablo:
> use strict version

Those change will be override by #1719

parmentf (approved):

parmentf (approved):

touv (approved):

PR #1718, opened 2023-10-03
check the first object before generate tar file

touv:


touv:
For example, if the chosen field does not exist we get this error


  
  ```bash
curl -v 'http://localhost:31976/exporters/bundle' \
       -H 'accept: application/json' \
       -H 'Content-Type: application/json' \
       -d '{"field": ["toto"], "connectionStringURI": "mongodb://mongo:27017/lodex", "host": "http://localhost:3000"}'

```

image

JulienMattiussi (approved):

PR #1734, opened 2023-10-17
Fix bar chart sort

AlasDiablo:


src/app/js/formats/vega-lite/models/BarChart.js
```diff
@@ -187,26 +174,10 @@ class BarChart extends BasicChart {
...
-                    this.direction === AXIS_VERTICAL ? 'x' : 'y';
-                break;
-            case LABEL_DESC:
-                labelsModel.encoding.y.sort = model.encoding.x.sort =
-                    this.direction === AXIS_VERTICAL ? '-x' : '-y';
-                break;
-            default:
-                break;
-        }
+        labelsModel.encoding.x.sort = null;
```
parmentf, line 180: Add a comment to say that you disable vega's sorts, because they're already done by routines (and because, when both are enabled, the sort was wrong, inexplicably).

parmentf (changes requested):

parmentf (approved):

touv (approved):

PR #1740, opened 2023-10-19
Update webpack

AlasDiablo:
La derniere version de webpack affiche les erreur il y aura donc quelle que message revenent souvant comme :
- The requested module 'd3-collection' contains conflicting star exports for the name 'map' with the previous requested module 'd3-array'
- ResizeObserver loop completed with undelivered notifications
Ces deux message sont non bloquant l'un a un fixe ici https://github.com/Inist-CNRS/lodex/pull/1751


package.json
```diff
@@ -95,19 +95,19 @@
         "@uswitch/koa-tracer": "1.7.2",
         "ace-builds": "1.4.14",
         "aphrodite": "^1.2.5",
-        "babel-loader": "^8.2.4",
+        "babel-loader": "9.1.3",
```
arimet, line 98: It's intended to fix versions and prevent patches ?

AlasDiablo: Technically yes, I follow [npmrc](https://github.com/Inist-CNRS/lodex/blob/master/.npmrc) rule

touv (approved):

parmentf (approved):

PR #1762, opened 2023-11-07
Prevent syntax error in customized loaders

parmentf:
When using loaders with ezs multiline statements in customized loader and modifying it, a syntax error happens.

image

https://github.com/Inist-CNRS/lodex/blob/b68414f31f667cd6e3d3b795dc283b3019823569/workers/loaders/zip.ini#L113-L115

I think that ace editor modifies its content on change (something like converting LF line returns to CR LF).


parmentf:
See <https://mattermost.inist.fr/projets/pl/ozs17gtodjnmbprx1quwoqj8dc>

PR #1767, opened 2023-11-09
Limit size of tenant name to 50

JulienMattiussi:
image


arimet:
Maybe add an information about the limit in the form

arimet (approved):

PR #1768, opened 2023-11-10
Dynamic theme system

AlasDiablo:
- I got some error on the Admin config page
- Themes are missing some data and use vulnerable library (bad file import or theme not compatible ?)
- Themes need to be refactor to avoid getting error due to bad css and js

TODO 
- [x] Cleanup themes
- [x] Add new simple themes to test the poc
- [x] Refactor the frontend hydration process to eliminate the need for waiting on the theme, and instead utilise the default option during the wait


src/api/controller/front.js
```diff
@@ -28,19 +28,14 @@ import configureStoreServer from '../../app/js/configureStoreServer';
 import Routes from '../../app/js/public/Routes';
 import translations from '../services/translations';
 import getLocale from '../../common/getLocale';
-import customTheme from '../../app/custom/customTheme';
 
 import { getPublication } from './api/publication';
 import getCatalogFromArray from '../../common/fields/getCatalogFromArray.js';
 import { DEFAULT_TENANT } from '../../common/tools/tenantTools';
+import getTheme, { getAvailableTheme } from '../models/themes';
```
parmentf: Shouldn't `Theme` be a plural, as you get all available theme**s** ?

src/api/models/themes.js
```diff
@@ -0,0 +1,131 @@
...
+const REGEX_THEMES_HOST = /\{\|__THEMES_HOST__\|\}/g;
+
+/**
+ * @type {Map<string, {name: {fr: string, en: string}, description: {fr: string, en: string}, index: string, customTheme: *}>}
+ */
+const loadedThemes = new Map();
+/**
+ * @type {Map<string, boolean>}
+ */
+const availableTheme = new Map();
```
parmentf: Should be `availableThemes`.

src/api/controller/front.js
```diff
@@ -343,6 +348,15 @@ app.use(
     }),
 );
 
+for (let theme of getAvailableTheme()) {
+    app.use(
+        mount(
+            `/themes/${theme}/`,
+            serve(path.resolve(__dirname, `../../themes/${theme}`)),
+        ),
```
parmentf, line 421: It seems that when you mount a theme, you can't specialize any files of the theme, or simply add files (from `custom` directory) any more.
Am I wrong?

AlasDiablo: custom directory still work, but this mount the theme directory to get acces to theme files

src/api/controller/front.js
```diff
@@ -214,10 +270,15 @@ const handleRender = async (ctx, next) => {
         initialEntries: [url],
     });
 
-    const theme = createTheme(customTheme, {
+    const configTenant = await ctx.configTenantCollection.findLast();
```
parmentf: Why do you take the last tenant's configuration?
Wouldn't it be better to use an instance's name (`default` for example, to get the classical theme)?

src/api/controller/front.js
```diff
@@ -214,10 +270,15 @@ const handleRender = async (ctx, next) => {
...
+    const lodexTheme = getTheme(configTenant.theme);
```
parmentf: Not sure to understand what `lodexTheme` is.
Is it the legacy custom theme? (the one once named `classic`)
Or is it any theme, from the config.json file?

src/api/models/themes.js
```diff
@@ -0,0 +1,148 @@
...
+ * @returns {{name: {fr: string, en: string}, description: {fr: string, en: string}, index: string, customTheme: *}}
+ */
+export const getTheme = (theme = 'default') => {
+    return loadedThemes.get(theme);
+};
+
+/**
+ * @returns {*[]}
+ */
+export const getAvailableThemes = () => {
```
parmentf, line 34: The whole function could have been a map:

```js
export const getAvailableThemes = () => loadedThemes.map((theme, value) => ({
  value,
  name: theme.name,
  description: theme.description
}));
```

This is only a refactor, and depends on your preferences.

src/api/models/themes.js
```diff
@@ -0,0 +1,148 @@
...
+
+const REGEX_JS_HOST = /\{\|__JS_HOST__\|}/g;
+const REGEX_THEMES_HOST = /\{\|__THEMES_HOST__\|}/g;
+
+/**
+ * @type {Map<string, {name: {fr: string, en: string}, description: {fr: string, en: string}, index: string, customTheme: *}>}
+ */
+const loadedThemes = new Map();
+/**
+ * @type {Map<string, boolean>}
```
parmentf, line 19: What does the boolean represent?
Availability?

src/api/models/themes.js
```diff
@@ -0,0 +1,148 @@
...
+                );
+            }
+
+            let indexLocation = '../../app/custom/index.html';
+            if (themeConfig?.files?.theme?.index) {
+                indexLocation = `../../${uri}/${themeConfig?.files?.theme?.index}`;
+            }
+
+            const index = fs
+                .readFileSync(path.resolve(__dirname, indexLocation))
```
parmentf: I know it's easier to write with `.readFileSync`, but it's blocking any other i/o at the same time (moreover, `index.html` could be a long file).
By default, I'd prefer the use of the async `.readFile`.
Of course, it makes that heavier to write.

src/api/models/themes.js
```diff
@@ -0,0 +1,148 @@
...
+                    `The ${theme} theme version may not be compatible with the current Lodex version. Expected theme version: 1, current version: ${themeConfig.version}.`,
+                );
+                logger.warn(
+                    'Lodex may attempt to load this theme. Please note that if you use this theme, it may cause visual problems.',
+                );
+            }
+
+            let indexLocation = '../../app/custom/index.html';
+            if (themeConfig?.files?.theme?.index) {
+                indexLocation = `../../${uri}/${themeConfig?.files?.theme?.index}`;
```
parmentf:
```suggestion
                indexLocation = `../../${uri}/${themeConfig.files.theme.index}`;
```

src/api/models/themes.js
```diff
@@ -0,0 +1,148 @@
...
+                .readFileSync(path.resolve(__dirname, indexLocation))
+                .toString()
+                .replace(REGEX_JS_HOST, jsHost)
+                .replace(REGEX_THEMES_HOST, themesHost);
+
+            let customTheme = defaultCustomTheme;
+            if (themeConfig?.files?.theme?.main) {
+                customTheme = (
+                    await import(
+                        `../../${uri}/${themeConfig?.files?.theme?.main}`
```
parmentf:
```suggestion
                        `../../${uri}/${themeConfig.files.theme.main}`
```

src/api/models/themes.js
```diff
@@ -0,0 +1,148 @@
...
+            const index = fs
+                .readFileSync(path.resolve(__dirname, indexLocation))
+                .toString()
+                .replace(REGEX_JS_HOST, jsHost)
+                .replace(REGEX_THEMES_HOST, themesHost);
+
+            let customTheme = defaultCustomTheme;
+            if (themeConfig?.files?.theme?.main) {
+                customTheme = (
+                    await import(
```
parmentf: Rewrite fields, to avoid cases where fields are non-existing.

parmentf (changes requested):

src/api/controller/front.js
AlasDiablo: This variable contain the theme use by the tenant

src/api/controller/front.js
AlasDiablo: I reuse some code from the tenant config, and in this collection we have only one element

src/api/models/themes.js
AlasDiablo: If the theme was successfully loaded

src/app/js/admin/configTenant/ConfigTenantForm.js
```diff
@@ -147,6 +179,21 @@ export const ConfigTenantForm = ({
...
+                    setTheme(event.target.value);
+                }}
+            >
+                {themes.map(t => (
+                    <MenuItem key={t.value} value={t.value}>
+                        {t.name[polyglot.currentLocale]} -{' '}
+                        {t.description[polyglot.currentLocale]}
+                    </MenuItem>
+                ))}
+            </Select>
```
AlasDiablo: Mettre une taille fixe

parmentf (approved):

src/api/controller/front.js
arimet: Could you use ctx.configTenant here ?

touv:
@parmentf je pense que l'on peut merger cetet PR quand @AlasDiablo aura résolu les conflits. 

Les problèmes d'affichage dans les thèmes ne concernent cette pr et pourront faire l'objet d'autres corrections

PR #1799, opened 2023-11-24
To avoid sending an unauthorized header to api.istex.fr

touv:


touv:
https://trello.com/c/9raa3NV9/507-versions-14-format-requ%C3%AAte-istex-ne-fonctionne-plus

arimet (approved):

PR #1815, opened 2023-11-30
Use parse int and also get the value and not the html element

AlasDiablo:


parmentf:
Do you use `parseInt()` for a particular usage?
I prefer using `Number()`, as it has only one parameter, contrary to `parseInt()`, which can lead to errors when used within a `map`.
However, `parseInt` can be practical when you want to parse a string beginning with a number.

```js
parseInt("12kjhiut") // returns 12
```

parmentf (approved):

PR #1836, opened 2023-12-13
Fix parallel coordinates chart

AlasDiablo:


workers/routines/distance-with.ini
```diff
@@ -120,16 +121,17 @@ than = env('minValue')
...
-[tune]
-path = env('tuneBy', 'value')
-
-; On tri sur le champ value
-[sort]
-reverse = env('reverse')
-
-; On garde uniqument les valeurs (donc on supprime les informations qui ont permis le tri)
-[value]
+; This part is disable due to a bug on "value" instruction
```
touv: maybe an idea to fix the problem rather than ignoring it

touv (changes requested):

parmentf (changes requested):
There was a mistyping in `src/app/js/formats/parallel-coordinates/ParallelCoordiantesChartView.js`, which should be `src/app/js/formats/parallel-coordinates/ParallelCoordinatesChartView.js` (`a` and `n` inverted)

PR #1839, opened 2023-12-15
use mongo instead of files

touv:


docker-compose.dev.yml
```diff
@@ -81,6 +81,7 @@ services:
             EZS_CONTINUE_DELAY: '60'
             EZS_NSHARDS: '31'
             EZS_CACHE: 'false'
+            EZS_VERBOS: 'true'
```
parmentf: Isn't it `EZS_VERBOSE` (with a final `E`)?

src/app/custom/translations.tsv
```diff
@@ -698,7 +698,7 @@
 "enrichment_logs"	"Enrichment logs"	"Logs de l'enrichissement"
 "precomputed_logs"	"Precomputed data logs"	"Logs des données précalculées"
 "precomputed_data"	"Precomputed data"	"Données précalculées"
-"precomputed_data_limit"	"We limit the sample to 600 chars"	"Nous limitons l'échantillon à 600 caractères"
+"precomputed_data_limit"	"We limit the sample to the first 10 results"	"Nous limitons l'échantillon au 10 premiers résulats"
```
parmentf:
```suggestion
"precomputed_data_limit"	"We limit the sample to the first 10 results"	"Nous limitons l'échantillon aux 10 premiers résultats"
```

docker-compose.dev.yml
touv: I understand why it didn't work ;-)

PR #1851, opened 2024-01-12
Split text and routine data input

AlasDiablo:


src/app/js/fields/sourceValue/SourceValueToggle.js
```diff
@@ -40,6 +42,20 @@ const TRANSFORMERS_FORM_STATUS = new Map([
             },
         ],
     ],
+    [
+        'arbitrary',
```
touv: @AlasDiablo  why not use the keyword “routine”, that would certainly be more meaningful.

src/app/js/fields/transformers/TransformerList.js
```diff
@@ -51,9 +51,12 @@ const getHiddenTransformers = source => {
     if (source === 'fromColumnsForSubRessource') {
         return 3;
     }
-    if (source === 'fromColumns' || source === 'arbitrary') {
+    if (source === 'fromColumns' || source === 'arbitrary_value') {
         return 1;
     }
+    if (source === 'arbitrary') {
```
touv: see my comment in the file above

touv (changes requested):

src/app/js/fields/sourceValue/SourceValueToggle.js
AlasDiablo: I have made this change for backwards compatibility, but upon further testing, it is not used in the models and can be renamed.

touv (approved):

PR #1859, opened 2024-01-19
refactor: move jsonallvalue first draft

AlasDiablo:
- [x] make the source code more abstract
- [x] founds better name for the folder and api route
- [ ] test the modification with an adapted dataset and model
  - [x] Manual call of the routine
  - [ ] Test with the format who required this routine


AlasDiablo:
For information i don't have a dataset and models to use the routine jsonallvalue, but with manual call it still work correctly and return all values.

touv (commented):
Why is a "first draft" ?
Is it ready for production ?

PR #1867, opened 2024-01-29
Fix: pagination lost in data grid

adguernier:
After navigate in pagination, the page number become `undefined` if the we change total item number


slax57 (commented):
It works fine and solves the issue.

I'm just wondering if this implementation isn't a bit too 'naive', and we should be a bit more clever when selecting the new page to land on.

Let's wait for feedback.

slax57 (approved):
Turns out we can keep this behavior :slightly_smiling_face:

PR #1871, opened 2024-01-30
Fix datagrid desynchronization between rows and header

slax57:
### Problems

1. Datagrid rows and headers can become misaligned when hiding/showing columns

https://github.com/Inist-CNRS/lodex/assets/14542336/a4d96688-3a06-4f0e-a1d1-2c9b58d47194

2. Datagrid rows and headers can become misaligned after clicking on a header field label and then scrolling horizontally

https://github.com/Inist-CNRS/lodex/assets/14542336/cda29299-7e5d-4586-bd55-80d7c4568fdc

### Solutions

1. This issue comes from https://github.com/mui/mui-x/issues/3890 , which has been fixed in https://github.com/mui/mui-x/releases/tag/v5.6.0 . This PR updates `@mui/x-data-grid` to `5.8.0` so this fix is included
2. This issue comes from https://github.com/mui/mui-x/issues/4275 , which has been fixed in https://github.com/mui/mui-x/releases/tag/v5.9.0 . Unfortunately, `@mui/x-data-grid@5.9.0` no longer works with React 16, so it requires updating to React 17 or 18 to be fixed.
- Workaround: clicking 'Hide all' and 'Show all' allows to reset the display in case this issue happens

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
                 "@fortawesome/react-fontawesome": "^0.1.18",
                 "@mui/icons-material": "5.14.19",
                 "@mui/material": "5.3.1",
-                "@mui/x-data-grid": "5.6.1",
+                "@mui/x-data-grid": "5.8.0",
                 "@recuperateur/async-busboy": "file:./packages/@recuperateur/async-busboy",
                 "@recuperateur/resumablejs": "file:./packages/@recuperateur/resumablejs",
                 "@uswitch/koa-access": "2.9.2",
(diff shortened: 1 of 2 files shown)


adguernier (approved):

bondi155:
Hello , is my first time here , sorry if this is not the place , i read that bug fixed in 5.60v but i have @mui/x-data-grid": "^6.9.1 and react 18 and i still have the same problem...( Datagrid rows and headers can become misaligned after clicking on a header field label and then scrolling horizontally) ...

PR #1876, opened 2024-01-31
Feat: keep hidden resources on re-publishing

adguernier:
## Problem
Unpublishing data results in the loss of previously hidden resources, which is annoying because users have to hide the resources again after another publication.

## Solution
Keep hidden resources and apply them during re-publishing

### DIFF
--- a/src/api/controller/api/publishedDataset.js
+++ b/src/api/controller/api/publishedDataset.js
@@ -93,13 +93,13 @@ export const getRemovedPage = async ctx => {
 
 export const removeResource = async ctx => {
     const { uri, reason } = ctx.request.body;
-    const date = new Date();
+    const removedAt = new Date();
     ctx.hiddenResource.create({
         uri,
         reason,
-        date,
+        removedAt,
     });
-    ctx.body = await ctx.publishedDataset.hide(uri, reason, date);
+    ctx.body = await ctx.publishedDataset.hide(uri, reason, removedAt);
 };
 
 export const restoreResource = async ctx => {
--- a/src/api/controller/api/publishedDataset.js
+++ b/src/api/controller/api/publishedDataset.js
@@ -94,7 +94,7 @@ export const getRemovedPage = async ctx => {
 export const removeResource = async ctx => {
     const { uri, reason } = ctx.request.body;
     const removedAt = ctx.request.body.removedAt ?? new Date();
-    ctx.hiddenResource.create({
+    await ctx.hiddenResource.create({
         uri,
         reason,
         removedAt,
@@ -104,7 +104,7 @@ export const removeResource = async ctx => {
 
 export const restoreResource = async ctx => {
     const { uri } = ctx.request.body;
-    ctx.hiddenResource.deleteByUri(uri);
+    await ctx.hiddenResource.deleteByUri(uri);
     ctx.body = await ctx.publishedDataset.restore(uri);
 };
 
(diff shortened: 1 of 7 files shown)


src/api/controller/api/publishedDataset.js
```diff
@@ -93,12 +93,18 @@ export const getRemovedPage = async ctx => {
 
 export const removeResource = async ctx => {
     const { uri, reason } = ctx.request.body;
-
-    ctx.body = await ctx.publishedDataset.hide(uri, reason);
+    const removedAt = ctx.request.body.removedAt ?? new Date();
+    ctx.hiddenResource.create({
```
slax57: **question:** ‏Shouldn't you _await_ this call?

```suggestion
    await ctx.hiddenResource.create({
```

src/api/controller/api/publishedDataset.js
```diff
@@ -93,12 +93,18 @@ export const getRemovedPage = async ctx => {
...
+        uri,
+        reason,
+        removedAt,
+    });
+    ctx.body = await ctx.publishedDataset.hide(uri, reason, removedAt);
 };
 
 export const restoreResource = async ctx => {
     const { uri } = ctx.request.body;
+    ctx.hiddenResource.deleteByUri(uri);
```
slax57: **question**: ‏Same, shouldn't you await this call?

```suggestion
    await ctx.hiddenResource.deleteByUri(uri);
```

src/api/services/publishDocuments.js
```diff
@@ -26,6 +30,7 @@ export const versionTransformerDecorator = (
                 publicationDate,
             },
         ],
+        ...hiddenResource,
```
slax57: **question:** ‏Why not return `hiddenResource` instead of `...hiddenResource`? I'm worried some fields in `hiddenResource` could overlap with other fields from the object, like `uri` for instance...

src/api/services/publishDocuments.spec.js
```diff
@@ -45,6 +45,9 @@ const getCtx = ({ subresources } = {}) => ({
     publishedDataset: {
         insertBatch: 'publishedDataset.insertBatch()',
     },
+    hiddenResource: {
+        findAll: jest.fn().mockImplementation(() => null),
```
slax57: **nitpick (non-blocking):** ‏You should probably return an empty array here instead of null, to make the mocked implementation more realistic

slax57 (commented):
Also, remember to change the PR name ;)

src/api/services/publishDocuments.js
slax57: Actually I get why you need to 'merge' the properties in the object (to populate the `removedAt` field at least), but my concern about the `uri` field still applies

adguernier: Resolved by moving `...hiddenResource` at the top

src/api/services/publishDocuments.spec.js
```diff
@@ -222,6 +227,50 @@ describe('publishDocuments', () => {
...
+                    uri: 'uid:/09P2JFN2',
+                    reason: 'reason',
+                    removedAt: date,
+                },
+                {
+                    uri: 'uid:/0R4JCK4F',
+                    reason: 'reason',
+                    removedAt: date,
+                },
+            ];
```
slax57: **nitpick (non-blocking):** ‏You could use two different values for `reason`, which would allow to check that the one you get in the result is indeed the one matching the resource uri.

slax57 (commented):
Other than that, LGTM!

slax57 (approved):

PR #1880, opened 2024-02-02
Feat export hidden resources

adguernier:
This PR attempts to add a way to export hidden resources as JSON

### DIFF
--- a/src/api/controller/api/hiddenResource.js
+++ b/src/api/controller/api/hiddenResource.js
@@ -7,18 +7,12 @@ const app = new Koa();
 
 const getExportHiddenResources = async ctx => {
     const hiddenResources = await ctx.hiddenResource.findAll();
-    const filePath = `/tmp/hiddenResources.json`;
+    const date = new Date().getTime();
+    const filePath = `/tmp/hiddenResources${date}.json`;
     const stream = fs.createWriteStream(filePath);
     stream.write(JSON.stringify(hiddenResources));
     stream.end();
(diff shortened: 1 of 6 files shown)


src/api/controller/api/hiddenResource.js
```diff
@@ -0,0 +1,22 @@
...
+    const hiddenResources = await ctx.hiddenResource.findAll();
+    const date = new Date().getTime();
+    const filePath = `/tmp/hiddenResources${date}.json`;
+    const stream = fs.createWriteStream(filePath);
+    stream.write(JSON.stringify(hiddenResources));
+    stream.end();
+    stream.on('finish', () => unlinkFile(filePath));
+    ctx.body = fs.createReadStream(filePath);
+    ctx.status = 200;
+};
```
slax57: **issue:** ‏We should wait for the stream to finish writing before returning a 200 status code. Currently there is no link between the Promise resolution status of the function and the stream completion.

Looking at existing code, I found a helper function called `streamEnd`, used in `src/api/controller/api/dump.js`, that could probably help here.

EDIT: Actually, looking at this again, I don't even understand why you need to write into a file in the first place. Can't you simply feed `ctx.body` with `JSON.stringify(hiddenResources)`? :thinking:

slax57 (changes requested):
Other than that it looks good to me :slightly_smiling_face: 

I assume we will implement the import feature in another PR?

slax57 (approved):

PR #1883, opened 2024-02-02
Fix precomputed job cancelation

slax57:
### Problems

- The 'CANCEL' button in the precomputed form is misleading, as we could think it allows to cancel the job, while it really only allows to cancel the edition of the precomputed settings
- Clicking the cancel button from the progress bar for a precomputed job while in the ON_HOLD state does not seem to do anything
- Starting a precomputed job multiple times may raise errors when the results of the computation arrive late
- Deleting a precomputed job sometimes throws an error: `MongoServerError: ns not found`
- A progress bar is still visible after we have deleted a precomputed job in the ON_HOLD state

### Solutions

- Rename the 'CANCEL' button to 'BACK' to make it clearer what it does
- Do not display the cancel button in the progress bar for precomputed jobs: indeed, such jobs must be DELETED to cancel them
- Fix errors when Deleting a precomputed job
- Properly remove the progress bar when a precomputed job is deleted, regardless of its state
- Detect when we receive the computation results from the webservice for a precomputed job that was deleted, and simply do nothing in this case

### Testing

This PR includes a [commit](https://github.com/Inist-CNRS/lodex/commit/c4be06fdad71bef6a7df9c73dff712315f9dfba8) that was reverted, which contains some code allowing to test the different status of a precomputed job.

To use it, re-apply the changes from that commit, then run `make start-dev` again.

Once it's done, you will be able to create a precomputed job with the following URL: http://json-server:3001/webhook

This fake service is set up to answer to the precomputed job after 20 seconds, and then call the success callback after 20 more seconds. Timers can be configured in the `server-json.js` file.


src/api/controller/api/precomputed.js
```diff
@@ -65,13 +65,8 @@ export const putPrecomputed = async (ctx, id) => {
...
     try {
-        const activeJob = await getActiveJob(ctx.tenant);
-        if (
-            activeJob?.data?.jobType === PRECOMPUTER &&
-            activeJob?.data?.id === id
-        ) {
-            cancelJob(ctx, PRECOMPUTER);
-        }
+        const precomputed = await ctx.precomputed.findOneById(id);
+        await cancelJob(ctx, PRECOMPUTER, precomputed?.name);
```
slax57, line 69: Calling `cancelJob` is also useful when there is a completed job for this precomputed task, not only when there is an active job. So we should call it all the time and let the `cancelJob` function decide what needs to be done.

src/api/models/precomputed.js
```diff
@@ -29,7 +29,12 @@ export default async db => {
     };
 
     collection.delete = async id => {
-        await db.collection(`pc_${id}`).drop();
+        try {
+            await db.collection(`pc_${id}`).drop();
+        } catch {
+            // Collection does not exist, no big deal
+            console.warn(`Failed to drop collection 'pc_${id}'`);
+        }
```
slax57, line 37: This circumvents `MongoServerError: ns not found` errors I got, which seem to be consecutive to https://github.com/Inist-CNRS/lodex/pull/1860. @touv feel free to suggest a better implementation if you disagree with this one! :grin:

adguernier (approved):

PR #1893, opened 2024-02-13
Add the possibility to delete the database

AlasDiablo:


src/api/controller/rootAdmin.js
```diff
@@ -131,8 +131,12 @@ const deleteTenant = async ctx => {
         ctx.status = 403;
         ctx.body = { error: `Invalid name: "${name}"` };
     } else {
-        deleteWorkerQueue(tenantExists.name);
+        deleteWorkerQueue(tenantExists.name).then();
```
parmentf, line 134: **question:** ‏What?

I don't get the utility of this `.then()`.
OK it shows that `deleteWorkerQueue` is asynchronous.

I just hope it's not just an unachieved development.

AlasDiablo: This is just to remove warring, this function is call, be we don't use the result.
We just wan to run-it

parmentf (approved):

PR #1941, opened 2024-04-08
Fix tenant name in page title to avoid crash on server side

AlasDiablo:


src/app/js/lib/getTitle.js
```diff
@@ -1,13 +1,26 @@
...
-     * @type {string}
+     * @type {string[]}
      */
-    const tenant = window.__TENANT__ ?? 'default';
+    const title = [];
+
+    if (window && window.__TENANT__) {
+        title.push(window.__TENANT__);
+    } else {
+        title.push('default');
```
parmentf: **question:** Does it mean that on server side, the name of the tenant will always be `default` ?

Even when not on the `default` tenant.
Wouldn't it be better to set another name  (best: put the real tenant name).‏

src/app/js/lib/getTitle.js
```diff
@@ -1,13 +1,26 @@
...
+    }
+
+    if (host) {
+        title.push(HOST_REGEX.exec(host)[1]);
+    } else {
+        title.push('example');
+    }
```
parmentf: **question:** ‏Why display `example` when you don't have the host?
```suggestion
```

AlasDiablo: Don't no, this was this way before.

parmentf (approved):

PR #2023, opened 2024-06-11
Add a button to go to admin page from the root-admin

AlasDiablo:
Screenshot_20240611_134941

### DIFF
--- a/src/app/js/root-admin/Tenants.js
+++ b/src/app/js/root-admin/Tenants.js
@@ -17,7 +17,15 @@ import {
     GridToolbarDensitySelector,
     GridToolbarFilterButton,
 } from '@mui/x-data-grid';
-import { Button, Link, Tooltip, Typography } from '@mui/material';
+import {
+    Box,
+    Button,
+    IconButton,
+    Link,
+    Tooltip,
+    Typography,
+} from '@mui/material';
+import AdminPanelSettingsIcon from '@mui/icons-material/AdminPanelSettings';
 
 const baseUrl = getHost();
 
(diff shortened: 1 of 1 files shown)


src/app/js/root-admin/Tenants.js
```diff
@@ -378,12 +409,13 @@ const Tenants = ({ handleLogout }) => {
             flex: 1,
             renderCell: (params) => {
                 return (
-                    <Button
+                    <IconButton
```
parmentf, line 412: **question:** ‏What's the difference between `Button` and `IconButton` ?

src/app/js/root-admin/Tenants.js
```diff
@@ -267,15 +270,43 @@ const Tenants = ({ handleLogout }) => {
...
+                                '&:hover': {
+                                    textDecoration: 'none',
+                                },
+                            }}
+                        >
+                            {params.value}
+                        </Link>
+                        <IconButton
+                            size="large"
+                            href={`${baseUrl}/instance/${name}/admin`}
```
parmentf, line 303: **praise:** ‏Good idea!

parmentf (approved):

src/app/js/root-admin/Tenants.js
AlasDiablo: Is a variant of the Button when we have one icon as the children.
For example when you want a button with one of the following icon https://mui.com/material-ui/material-icons/

AlasDiablo: You can learn for here:
https://mui.com/material-ui/react-button/#icon-button

PR #2034, opened 2024-06-14
Update treemap icon to always take the correct font size

AlasDiablo:


src/app/js/formats/vega/component/tree-map/TreeMapIcon.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import ViewQuiltIcon from '@mui/icons-material/ViewQuilt';
+
+const TreeMapIcon = (props) => (
+    // We size the fontSize to empty string to disable it
```
parmentf, line 5: **praise:** ‏Thanks for the comment, otherwise it would only be a silent workaround

parmentf (approved):

PR #2090, opened 2024-07-10
Root panel improvement

AlasDiablo:
***System Load indication***

Screenshot_20240710_114711

***Create tenant new gui***

Screenshot_20240711_084030

***Delete tenant new gui***

Screenshot_20240711_085621

***Update tenant new gui**

Screenshot_20240711_090911

### Todo

- [x] Update form to allow validation via keyboard


src/app/js/root-admin/SystemLoad.js
```diff
@@ -0,0 +1,246 @@
...
+            .then((response) => response.json())
+            .then((data) => {
+                setLoadTitle(
+                    `${data.loadavg[0]}% / 1 min, ${data.loadavg[1]}% / 5 min, ${data.loadavg[2]}% / 15 min`,
+                );
+                setLoadAvg(data.loadavg[0]);
+
+                const memPercent =
+                    ((data.totalmem - data.freemem) / data.totalmem) * 100;
+                const totalMen = sizeConverter(data.totalmem / 1024);
```
parmentf: **typo:** ‏Rename `totalMen` to `totalMem`

src/api/controller/rootAdmin.js
```diff
@@ -136,11 +137,26 @@ const deleteTenant = async (ctx) => {
...
+const systemInfo = async (ctx) => {
+    const dbStats = await ctx.rootAdminDb.stats({ scale: 1024 });
+    ctx.body = {
+        database: {
+            total: dbStats.fsTotalSize,
+            use: dbStats.fsUsedSize,
+        },
+        loadavg: os.loadavg(),
+        totalmem: os.totalmem(),
+        freemem: os.freemem(),
```
parmentf, line 175: **issue:** ‏This is the same statement than in ezMaster, and this is not the right indicator.

ezMaster uses:
https://github.com/Inist-CNRS/ezmaster/blob/master/ezmaster-api/lib/event-refresh-infos-machine.js#L44

I don't know if `os` proposes a better indicator for the free memory, but all users think that the whole memory is used, although it is only reserved.

src/app/js/root-admin/CreateTenantDialog.js
```diff
@@ -1,94 +1,163 @@
...
+        /**
+         * @type {string}
+         */
+        const newName = event.target.value;
+        // We replace any accented and special char with the base letter or a dash
+        // https://stackoverflow.com/questions/70287406/how-to-replace-all-accented-characters-with-english-equivalents
+        // https://stackoverflow.com/questions/36557202/replacing-special-characters-with-dashes
+        const nameWithoutAccentedCharacters = newName
+            .normalize('NFD')
+            .replace(/[\u0300-\u036f]/g, '');
```
parmentf: **suggestion:** ‏Maybe it would be clearer to use [lodash's `deburr`](https://lodash.com/docs/4.17.15#deburr)

src/app/js/root-admin/CreateTenantDialog.js
```diff
@@ -1,94 +1,163 @@
...
+        const nameWithoutSpecialChar = nameWithoutAccentedCharacters
+            .replace(/\s+/g, '-')
+            .replace(/\W+(?!$)/g, '-');
+        let nameWithoutFirstLetterAsDash = nameWithoutSpecialChar;
+        if (nameWithoutSpecialChar.startsWith('-', 0)) {
+            nameWithoutFirstLetterAsDash = nameWithoutSpecialChar.substring(
+                1,
+                nameWithoutSpecialChar.length,
+            );
+        }
```
parmentf: **issue:** ‏And what about `nameWithoutSpecialChar` beginning with more than one dash ?

I'd replace `if` with `while`. And then check if `nameWithoutFirstLetterAsDash` is empty...

src/app/js/root-admin/CreateTenantDialog.js
```diff
@@ -1,94 +1,163 @@
...
+        const nameInLowerCase = nameWithoutFirstLetterAsDash.toLowerCase();
+        setName(nameInLowerCase);
```
parmentf: **issue:** ‏What about an empty `nameInLowerCase` ?

:thinking: 
Maybe it's not the place for checking this.

src/app/js/root-admin/rootAdminUtils.js
```diff
@@ -0,0 +1,17 @@
+export const sizeConverter = (value) => {
+    if (value == null) {
+        return '-';
+    }
+
+    const mbSize = (value / 1024).toFixed(2);
+
+    if (mbSize > 1024) {
+        return `${(mbSize / 1024).toFixed(2)} Gio`;
```
parmentf: **thought (non-blocking):** And what if you eventually need Tio‏ ?

I know it's too much to install a package only to do that, but you may find inspiration to be more generic.

src/api/controller/rootAdmin.js
AlasDiablo: We need to test thit with an working version on our infrastructure, beceaus in my docker env i don't have any issues, i have also tested with systeminformation and i have the same result.

src/app/js/root-admin/CreateTenantDialog.js
```diff
@@ -1,94 +1,156 @@
...
 } from '../../../common/tools/tenantTools';
 
+const cleanUpName = (name) => {
+    // We replace any accented and special char with the base letter or a dash
+    // https://stackoverflow.com/questions/36557202/replacing-special-characters-with-dashes
+
+    return deburr(name)
+        .replace(/\s+/g, '-')
+        .replace(/_/g, '-')
+        .replace(/\W+/g, '-')
```
parmentf: **suggestion:** You could gather these lines (only one regex to apply, don't know it's quicker or not).

```suggestion
        .replace(/[\s_\W]+/g, '-')
```

‏Other advantage: There will at most be one `-` at the beginning of the string.

src/app/js/root-admin/SystemLoad.js
```diff
@@ -0,0 +1,242 @@
...
+                setMemUsage(memPercent);
+                setMemTile(`Mémoire : ${useMem} / ${totalMem}`);
+
+                const storagePercent =
+                    (data.database.use / data.database.total) * 100;
+                const totalStorage = sizeConverter(data.database.total);
+                const useStorage = sizeConverter(data.database.use);
+
+                setStorageUsage(storagePercent);
+                setStorageTile(`Stockage : ${useStorage} / ${totalStorage}`);
```
parmentf: **suggestion:** ‏I find it clearer that way

I would have computed `usedMem` and `totalMem` first, then `memPercent = usedMem / totalMem * 100`.
Of course, `sizeConverter` would be called later.

Oh: and I multiply by 100 first, to lose less precision (but I'm not completely satisfied by the prior division by 1024).

```suggestion
                const totalMem = data.totalmem / 1024;
                const usedMem = (data.totalmem - data.freemem) / 1024;
                const memPercent = 100 * usedMem / totalMem;

                setMemUsage(memPercent);
                setMemTile(`Mémoire : ${sizeConverter(usedMem)} / ${sizeConverter(totalMem)}`);

                const storagePercent = 100 * data.database.use / data.database.total;
                const totalStorage = sizeConverter(data.database.total);
                const usedStorage = sizeConverter(data.database.use);

                setStorageUsage(storagePercent);
                setStorageTile(`Stockage : ${usedStorage} / ${totalStorage}`);
```

parmentf (approved):

src/app/js/root-admin/rootAdminUtils.js
```diff
@@ -1,17 +1,14 @@
+import { formatBytes } from '@draconides/format';
+
+const numberFormatter = new Intl.NumberFormat('fr-FR');
```
parmentf, line 3: **praise:** ‏Good idea!

parmentf (approved):

PR #2119, opened 2024-07-26
Update config.json

AnaelKremer:
By adding query-conditor-for-halcnrs loader


touv:
cf. https://github.com/Inist-CNRS/lodex/pull/2115

PR #2120, opened 2024-07-26
Update translations.tsv

AnaelKremer:
add transflations for the new loader "query-conditor-for-halcnrs"


touv:
cf. https://github.com/Inist-CNRS/lodex/pull/2115

PR #2124, opened 2024-08-02
to stop the recurcive loop

touv:


touv:
see https://github.com/Inist-CNRS/lodex/pull/2126

PR #2128, opened 2024-08-06
(1) upgrade mongodb driver

touv:
Attempts to make Lodex compatible with a mongodb cluster version 7


touv:
it may be necessary to install the latest version of jest to correct the problem https://github.com/jestjs/jest/releases/tag/v29.2.0

jonathanarnault:
Superseeded by v15

PR #2322, opened 2025-01-10
2315 reenable run on enrichment end

### LINKED ISSUE #2315: On ne peut plus lancer un enrichissement lorsqu'il s'est fini sans actualiser la page
Image

Désormais lorsque l'on lance un enrichissement et qu'on souhaite le modifier ou ajouter du code, on peut bien enregistrer les modifications, mais le bouton "lancer" se grise. Il convient de sortir de cet enrichissement pour le relancer soir par la liste soit en entrant de nouveau dans celui-ci.

ThieryMichel:
fix #2315

### DIFF
--- a/src/api/app.js
+++ b/src/api/app.js
@@ -178,4 +178,12 @@ app.on('uncaughtException', (error) => {
     console.error('Uncaught Exception', error);
 });
 
+process.on('unhandledRejection', (error) => {
+    console.error('Uncaught Rejection', error);
+});
+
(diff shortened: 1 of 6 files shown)


jonathanarnault (changes requested):
The button state works great but the enrichment state is not updated once it is finished as show in the video capture below


https://github.com/user-attachments/assets/65094359-582c-4686-bd19-c016d91100a6

jonathanarnault (approved):

PR #2359, opened 2025-01-16
Feat(ts): Init typescript support

jonathanarnault:


jonathanarnault:
Integrating typescript take smore time than expected, we skip migrating to typescript for now to focus on user features

PR #2362, opened 2025-01-16
V15 remove api test

ThieryMichel:
remove api test, that are broken with recent upgrades and are redundant with e2e tests


cypress/support/commands.js
jonathanarnault, line 1: Is this file still required ?

ThieryMichel: I was certain I had reverted that :p

jonathanarnault (approved):

PR #2375, opened 2025-01-20
Feat(annotation): Add backend services to support annotations

jonathanarnault:
Parts of #2324 

In this PR:
- Backend services

Missing from this PR:
- Admin
- Front support


src/api/controller/api/annotation.validator.js
```diff
@@ -0,0 +1,66 @@
+import { default as z } from 'zod';
+
+export const annotationSchema = z.object({
+    resourceId: z.string({
+        required_error: 'error_required',
+    }),
+    fieldId: z.string().nullish().default(null),
```
ThieryMichel: Didn't we say we would use a path instead, to support thing like pointing to a value inside an array of values ?

src/api/controller/api/annotation.validator.js
```diff
@@ -0,0 +1,66 @@
+import { default as z } from 'zod';
+
+export const annotationSchema = z.object({
+    resourceId: z.string({
+        required_error: 'error_required',
```
ThieryMichel: resourceId should be nullable to handle field on the dataset himself, as well as on the field itself regardless of the value.

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #2543, opened 2025-02-18
Fix translations and hover regressions on list view

### LINKED ISSUE #2539: Retirer le changement de couleur au survol des listes à puces
Fonctionnalité implantée lors de l'annotation d'une valeur et pas retirée
[Exemple](https://instance-globale-1510.lodex-dev.inist.fr/instance/demo-2025-02-19/uid:/09P2JFN2)

https://github.com/user-attachments/assets/68b825f0-afba-461a-a4f2-f3447a89eab9

### LINKED ISSUE #2540: Ne pas rendre filtrable la colonne de la checkbox
Retirer le filtre sur la colonne _#_checkbox_#_

https://github.com/user-attachments/assets/ef9109bb-6abe-49ee-9007-718bbe1ea2d6

### LINKED ISSUE #2541: Verbaliser les libellés des statuts d'annotations
Dans la datagrid des annotation les statuts sont en anglais :  `comment`, `addition`. Il faudrait changer par `commentaire`, `ajout`, `suppression`.

jonathanarnault:
Closes #2539 
Closes #2540
Closes #2541


src/app/custom/translations.tsv
```diff
@@ -1293,11 +1293,12 @@
 "annotation_add_content"	"Add some content"	"Ajouter du contenu"
 "removal"	"Removal"	"Suppression"
 "comment"	"Comment"	"Commentaire"
-"correction"	"Correction"	"Correction"
+"correct"	"Correction"	"Correction"
```
ThieryMichel, line 1297: Why?
We have either a removal, an addition, a comment, or a correction. A correct makes no sense IMO

ThieryMichel: Or if you want to keep "correct", then you have to rename "removal" to "remove" and "addition" to "add" to stay coherent

ThieryMichel (changes requested):

src/app/custom/translations.tsv
jonathanarnault: This is the name of the enum value in the validator

ThieryMichel: So you can fix the validator

jonathanarnault: Since we are using mongoDB, we do not have migration for existing values and this will probably break many existing annotations. Not sure that renaming the annotation value have any interest in this case as the name is explicit enough imho

ThieryMichel (approved):

PR #2547, opened 2025-02-18
Fix(annotation): Remove duplicate correction filter

jonathanarnault:


ThieryMichel (changes requested):
How about just removing the incorrect "correct" value from annotation.validator?
We do not want anyone using that value anyway. 
And you can rename `correct` to `correction` when retrieving the annotations from the database, avoiding any issue.

jonathanarnault:
> How about just removing the incorrect "correct" value from annotation.validator? We do not want anyone using that value anyway. And you can rename `correct` to `correction` when retrieving the annotations from the database, avoiding any issue.

This would break imports for existing annotations if I remove the correct annotation from kind enum

ThieryMichel:
> > How about just removing the incorrect "correct" value from annotation.validator? We do not want anyone using that value anyway. And you can rename `correct` to `correction` when retrieving the annotations from the database, avoiding any issue.
> 
> This would break imports for existing annotations if I remove the correct annotation from kind enum

If you really want to avoid any breaking change, then have the import rename correct to correction

ThieryMichel (approved):

PR #2627, opened 2025-02-28
Fix(dataset): Do not show the delete filtered button if there is no row

### LINKED ISSUE #2612: Bouton pour supprimer les lignes filrées reste actif
Dans l'onglet données. Lorsqu'un filtre est appliqué le bouton `supprimer les lignes filtrées` reste actif, même si la datagrid est vide.

Cliquer sur le bouton provoque une erreur.

Image

- LODEX 15.3.0

jonathanarnault:
Closes #2612

### DIFF
--- a/cypress/e2e/phase_1/publication.cy.js
+++ b/cypress/e2e/phase_1/publication.cy.js
@@ -205,6 +205,7 @@ describe('Dataset Publication', () => {
                 ['"2"', '"Alain"', '"Chabat"', 'true'].join(''),
             );
         });
+
         it('should allow to delete all filtered rows', () => {
             menu.openAdvancedDrawer();
             menu.goToAdminDashboard();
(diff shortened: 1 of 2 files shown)


cypress/e2e/phase_1/publication.cy.js
```diff
@@ -233,6 +234,23 @@ describe('Dataset Publication', () => {
             cy.findByText('No rows').should('be.visible');
         });
 
+        it.only('should allow not display the "delete all filtered rows" if no rows match the filter', () => {
```
ThieryMichel: .only spotted

ThieryMichel:
```suggestion
        it('should allow not display the "delete all filtered rows" if no rows match the filter', () => {
```

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #2653, opened 2025-03-03
Fix(autocomplete): Do not allow new values if not free solo

### LINKED ISSUE #2652: Admin > configuration d'un champ : Quand la "proposition d'une nouvelle valeur" dans une liste est désactivée, il est quand même possible en front de soumettre une nouvelle valeur, alors que ça ne devrait pas

jonathanarnault:
Closes #2652

### DIFF
--- a/src/app/js/annotation/fields/ProposedValueFieldList.js
+++ b/src/app/js/annotation/fields/ProposedValueFieldList.js
@@ -24,6 +24,7 @@ export function ProposedValueFieldList({
                 label={label}
                 options={options}
                 required
+                freeSolo={supportsNewValues}
             />
         );
     }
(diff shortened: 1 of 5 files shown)


src/app/js/annotation/fields/ProposedValueFieldList.js
```diff
@@ -24,6 +24,7 @@ export function ProposedValueFieldList({
                 label={label}
                 options={options}
                 required
+                freeSolo={supportsNewValues}
```
ThieryMichel: freeSolo is not very explicit, how about:
```suggestion
                supportsNewValues={supportsNewValues}
```

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #2767, opened 2025-03-20
To avoid hiding lost columns at the bottom of the dataset

touv:


touv:
makes lodex very slow for large dataset

touv:
Another idea would be to count the number of columns just before inserting them into the database and add them to each document.
We could then sort on this column and then hide it.

PR #2892, opened 2025-09-24
Fix vega graph aspect ratio when no image format is provided

Madeorsk:
#2886 

This PR fixes the default aspect ratio of a custom Vega Lite graph when the aspect ratio is not defined (set to "No aspect ratio"), and its height is set to `"container"`. It also fixes the aspect ratio selector, which saved the translated label of "none" aspect ratio.

### DIFF
--- a/src/app/js/formats/utils/components/vega-lite-component/VegaLiteComponent.js
+++ b/src/app/js/formats/utils/components/vega-lite-component/VegaLiteComponent.js
@@ -58,76 +58,45 @@ function CustomActionVegaLite({
             throw new Error('Invalid data injection type');
     }
 
+    const vegaGraphElement = (
+        <Vega
+            style={
+                aspectRatio === ASPECT_RATIO_NONE
+                    ? {
+                          width: '100%',
+                          aspectRatio:
+                              // If no height is provided when the spec height is set to container,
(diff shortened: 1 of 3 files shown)


src/app/js/formats/utils/components/vega-lite-component/VegaLiteComponent.js
```diff
@@ -65,7 +66,15 @@ function CustomActionVegaLite({
...
-                                ? { width: '100%' }
+                                ? {
+                                      width: '100%',
+                                      aspectRatio:
+                                          // If no height is provided when the spec height is set to container,
+                                          // the graph height is very small. As a workaround we set a default aspect ratio of 2/1.
+                                          specHeight === 'container'
+                                              ? '2 / 1'
+                                              : undefined,
+                                  }
```
ThieryMichel: Since the style is the same for both, how about defining it only once ?

ThieryMichel (changes requested):

src/app/js/formats/utils/components/vega-lite-component/VegaLiteComponent.js
Madeorsk: I did it with the whole element as the rest is also the same.

ThieryMichel (approved):

PR #2928, opened 2025-09-30
React hook form precomputed

ThieryMichel:
see https://github.com/orgs/Inist-CNRS/projects/5/views/1?pane=issue&itemId=130641136&issue=Inist-CNRS%7Clodex%7C2898


src/app/js/admin/progress/reducer.ts
```diff
@@ -1,4 +1,5 @@
 import { createAction, handleActions } from 'redux-actions';
+// @ts-expect-error TS7016
```
Madeorsk: I think you can remove this now.

src/app/js/admin/precomputed/PrecomputedForm.tsx
```diff
@@ -298,7 +113,7 @@ export const PrecomputedForm = ({
         }
     };
 
-    const handleAddPrecomputed = async () => {
+    const handleAddPrecomputed = async (formValues: PreComputation) => {
         const res = await createPrecomputed(formValues);
```
Madeorsk, line 133: I think it would be better to update the types of API functions too. I'd put `PreComputation` type in API instead of the view.

src/app/js/admin/precomputed/PrecomputedForm.tsx
```diff
@@ -312,7 +127,7 @@ export const PrecomputedForm = ({
         }
     };
 
-    const handleUpdatePrecomputed = async () => {
+    const handleUpdatePrecomputed = async (formValues: PreComputation) => {
         const { data, ...precomputedDataToUpdate } = {
             ...initialValues,
             ...formValues,
```
Madeorsk, line 152: same here, you could update the API functions types.

src/app/js/admin/precomputed/RunButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const RunButton = ({
+    handleLaunchPrecomputed,
+    precomputedStatus,
+    translate,
+    variant = 'contained',
+}: {
+    handleLaunchPrecomputed: (event: MouseEvent) => void;
+    precomputedStatus: string;
+    translate: (key: string) => string;
```
Madeorsk: Use `useTranslate` instead of passing this prop.

src/app/js/admin/precomputed/index.ts
```diff
@@ -19,19 +28,23 @@ export const initialState = {
...
     {
         LOAD_PRECOMPUTED: (state) => ({
             ...state,
             loading: true,
             initialized: true,
         }),
-        LOAD_PRECOMPUTED_ERROR: (state, { payload: error }) => ({
+        // @ts-expect-error TS7006
+        LOAD_PRECOMPUTED_ERROR: (
+            state: PrecomputedState,
```
Madeorsk, line 41: This type can be inferred

src/app/js/lib/selectors.ts
```diff
@@ -9,12 +10,33 @@ export const createGlobalSelector = (getLocalState, selector) =>
...
+//     getLocalState,
+//     selectors,
+// ) =>
+//     Object.keys(selectors).reduce(
+//         (result, key) => ({
+//             ...result,
+//             [key]: createGlobalSelector(getLocalState, selectors[key]),
+//         }),
+//         {},
+//     );
```
Madeorsk: Did you forget to remove this?

src/app/js/lib/selectors.ts
```diff
@@ -9,12 +10,33 @@ export const createGlobalSelector = (getLocalState, selector) =>
...
+export const createGlobalSelectors = <
+    T extends Record<string, (localState: any, props?: any) => any>,
```
Madeorsk, line 14: What about something like that? I think it would improve type checking.

```suggestion
    T extends Record<string, (localState: LocalState, props?: Props) => any>,
```

src/app/js/lib/components/Catalog.tsx
```diff
@@ -174,7 +177,7 @@ export const EnrichmentCatalog = ({
                                         : 'outlined'
                                 }
                             >
-                                {translate(filter)}
+                                {filter}
```
Madeorsk: You need to keep the call to translate here, otherwise the i18n for this button is completely removed.

Madeorsk (changes requested):

src/app/js/lib/selectors.ts
ThieryMichel: It's too soon to type the global state
We first need to do all sub reducer

src/app/js/lib/components/Catalog.tsx
ThieryMichel: No, the string come from a json, and there was not a single type key  in this json this that had a translation. 
(It is maintained by non developper)
All this did was add dozen of missing translation error in the console

src/app/js/admin/precomputed/index.ts
ThieryMichel: Yes, but handleActions type is too opinonated and would force us to refactor all actions creator to change their type. And all payload would have a type which is the union of all payload anyway.

src/app/js/lib/selectors.ts
ThieryMichel: It has too many side effect, we need to left some part untyped, otherwise it never ends

src/app/js/admin/precomputed/PrecomputedForm.tsx
ThieryMichel: We need to restrain ourself on typing otherwise we will never finish anything

src/app/js/admin/precomputed/PrecomputedForm.tsx
ThieryMichel: Same, I already typed a lot, we cannot type everything or we will never be finished

src/app/js/lib/components/Catalog.tsx
Madeorsk: Well, you're right about the category filter for precomputed fields. However, this component is also used in enrichments, and in this case it makes some difference. See (with translations at the top, without translations at the bottom):

image

Madeorsk (approved):

PR #2935, opened 2025-10-02
Fix type check

ThieryMichel:


tsconfig.json
```diff
@@ -1,7 +1,30 @@
...
+        "strict": true,
+        "noUnusedLocals": true,
+        "noUnusedParameters": true,
+        "erasableSyntaxOnly": true,
+        "noFallthroughCasesInSwitch": true,
+        "noUncheckedSideEffectImports": true,
+        "noImplicitAny": true
+    },
+    "include": ["vite.config.ts", "src/api", "src/common", "src/app"],
+    "exclude": ["test-utils.tsx"]
```
Madeorsk: Otherwise "build" directories are checked and test-utils.tsx is already fixed so we can just include it.

```suggestion
    "include": ["vite.config.ts", "test-utils.tsx", "src/api", "src/common", "src/app"],
    "exclude": ["src/build", "src/app/build"]
```

Madeorsk (changes requested):

Madeorsk (approved):

PR #2994, opened 2025-10-14
Migrate workers and ezsLodex to TypeScript

Madeorsk:
#2932

eszLodex wasn't compatible with tsx runtime because some JS functions passed to mongodb were transpiled. These functions are now defined in a JS string literal, so they won't be transpiled and we can use TypeScript in workers and ezsLodex.


packages/ezsLodex/src/convertToCSV.ts
```diff
@@ -0,0 +1,54 @@
+import omit from 'lodash/omit.js';
+
+// import { VALIDATED } from '../../common/propositionStatus'; // IN LODEX
```
ThieryMichel: commented out code

packages/ezsLodex/src/extractIstexQuery.ts
```diff
@@ -28,14 +28,14 @@ function removeNumberInstance(uri) {
  * }
  * @returns
  */
-export default function extractIstexQuery(data, feed) {
+export default function extractIstexQuery(this: any, data: any, feed: any) {
```
ThieryMichel, line 31: Better not to call an argument this, since it is a reserved keyword

packages/ezsLodex/src/extractIstexQuery.ts
```diff
@@ -1,4 +1,4 @@
-function removeNumberInstance(uri) {
+function removeNumberInstance(uri: any) {
```
ThieryMichel: uri should be string

packages/ezsLodex/src/aggregateQuery.ts
```diff
@@ -17,7 +18,7 @@ const stageNameTest = /^\$/;
  * @returns {Object}
  */
 export const createFunction = () =>
-    async function LodexAggregateQuery(data, feed) {
+    async function LodexAggregateQuery(this: any, data: any, feed: any) {
```
ThieryMichel, line 21: I'd rather add ts-expect-error, any silent way too many error down the code

ThieryMichel (changes requested):

packages/ezsLodex/src/aggregateQuery.ts
Madeorsk: I tried to use `@ts-expect-error` instead of any on this line, and it didn't show any silenced error. When you use `@ts-expect-error`, TypeScript still consider the param as any, it's just not as explicit.

Also, it has been made automatically (here and in other places) by `ts-migrate`, the same way it did in src/api.

packages/ezsLodex/src/extractIstexQuery.ts
Madeorsk: It's not an argument, it's a way to type hint `this` in this context. I know it's weird and unclear but it's how the code was before. We must keep it as is or change it completely.

(for reference, see https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function)

ThieryMichel: Oh Yeah, forgot that you could do that.

ThieryMichel (approved):

PR #2996, opened 2025-10-14
Fix search form inputs not updated properly after mutation

Madeorsk:
#2989

The form inputs were not updated properly after mutation. The fields stored internally were only updated after a full publication, and were _never_ updated when the dataset was not published.

### DIFF
--- a/src/app/js/fields/index.ts
+++ b/src/app/js/fields/index.ts
@@ -1,6 +1,11 @@
 import omit from 'lodash/omit';
 import uniq from 'lodash/uniq';
-import { combineActions, createAction, handleActions } from 'redux-actions';
+import {
+    combineActions,
+    createAction,
+    handleActions,
+    type ReducerMapValue,
+} from 'redux-actions';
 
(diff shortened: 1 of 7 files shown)


src/app/js/fields/index.ts
```diff
@@ -322,6 +336,83 @@ export default handleActions(
...
+            byName: Object.fromEntries(
+                Object.entries(state.byName).map(
+                    ([fieldName, field]: [string, Field]) => {
+                        if (field._id === _id) {
+                            return [fieldName, { ...field, overview }];
+                        }
+
+                        if (field.overview === overview) {
+                            const {
+                                overview: _overview,
```
ThieryMichel: Where does _overview come from ?

src/app/js/fields/index.ts
```diff
@@ -322,6 +336,83 @@ export default handleActions(
...
+                                isDefaultSortField: _isDefaultSortField,
```
ThieryMichel: same what is  _isDefaultSortField ?

ThieryMichel (approved):

PR #2998, opened 2025-10-14
Fix format admin fields

ThieryMichel:
#2977


src/app/js/formats/utils/components/field-set/FormatFieldSets.tsx
```diff
@@ -9,27 +7,30 @@ import AccordionSummary from '@mui/material/AccordionSummary';
...
-// @ts-expect-error TS7031
-const FormatFieldSet = ({ children, title, p, defaultExpanded, id }) => {
+type FormatFieldSetProps = {
+    children: ReactNode;
+    title: string;
+    defaultExpanded?: boolean;
+    id: string;
+};
+
+const VegaFieldSet = ({
```
Madeorsk: Should be named `FormatFieldSet`.

Madeorsk (changes requested):
The rest seems correct, but I fail to reproduce the bug that was reported in the first place.

Madeorsk (approved):

PR #3013, opened 2025-10-17
Split packages

Madeorsk:
#2880

- [X] `@lodex/common`
- [x] `@lodex/api`
- [x] `@lodex/workers`

To do in another PR after React upgrade:
- `@lodex/root-admin-app`
- `@lodex/admin-app`
- `@lodex/public-app`


packages/common/src/index.ts
ThieryMichel, line 1: Please don't add a barrel file. This is know to cause slow build.
see https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/

ThieryMichel (changes requested):
I do not see package.json in the tnew packages.
The goal is to setup a mono repo too.

packages/common/src/index.ts
Madeorsk: I know but for a library like I did, I'm forced to use at least one. I can remove the optional ones though.

ThieryMichel: Why ? Nothing prevent you to specify the full path while importing.

ThieryMichel (approved):

PR #3036, opened 2025-10-24
Search fields layout with subtitle

Madeorsk:
#2990

Following up our discussions, transform the third resource detail in a resource subtitle.

### DIFF
--- a/cypress/support/enrichmentFormPage.ts
+++ b/cypress/support/enrichmentFormPage.ts
@@ -43,6 +43,7 @@ export const updateNameEnrichment = () => {
     cy.contains('Save', { timeout: 500 }).click({
         force: true,
     });
+    cy.wait(3000);
     cy.reload();
     cy.findByLabelText('Name *', { timeout: 3500 }).should(
         'have.value',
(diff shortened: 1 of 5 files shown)


cypress/support/enrichmentFormPage.ts
```diff
@@ -43,6 +43,7 @@ export const updateNameEnrichment = () => {
     cy.contains('Save', { timeout: 500 }).click({
         force: true,
     });
+    cy.wait(3000);
```
ThieryMichel, line 46: Why did you need to add a wait here ?
Is it possible to detect a progressbar / notification instead?

ThieryMichel (changes requested):

cypress/support/enrichmentFormPage.ts
Madeorsk: > Why did you need to add a wait here ?

We need to wait for the save request to be fully done before reloading the page, otherwise the enrichment isn't saved yet when the page reload. I don't know why we didn't get it before.
Maybe I could track the notification or wait for the save button to be enabled again (it gets disabled when saving).

Madeorsk: I can't find a way to get it to work every time when I use the save button state... I don't really understand what's happening here, but with this small wait, the test always pass, and without it, it fails randomly.

ThieryMichel (approved):

PR #3080, opened 2025-11-07
Fix run button and pm2 config

ThieryMichel:


packages/admin-app/src/precomputed/RunButton.tsx
```diff
@@ -5,8 +5,8 @@ import PlayArrowIcon from '@mui/icons-material/PlayArrow';
 import { Button, type ButtonProps } from '@mui/material';
 import { type MouseEvent } from 'react';
 
-const RUNNABLE_STATUSES: (TaskStatusType | undefined)[] = [
-    undefined,
+const RUNNABLE_STATUSES: (TaskStatusType | '')[] = [
+    '',
```
jonathanarnault, line 10: I think we should support both empty string and undefined as precomputed are created with and undefined status when I test it locally

jonathanarnault (changes requested):

jonathanarnault (approved):

PR #3121, opened 2025-11-17
Advanced network

ThieryMichel:
[Screencast from 2025-11-17 11-23-12.webm](https://github.com/user-attachments/assets/50c76e9d-eeec-4c82-aee4-e37d434644c0)


packages/frontend-common/formats/chart/advancedNetwork/AdvancedNetworkView.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface NetworkProps {
+    colorSet?: string[];
+    formatData?: AdvancedNetworkData[];
+    field: Field;
+}
+
+const AdvancedNetwork = ({ formatData: _, colorSet, field }: NetworkProps) => {
+    const { nodes, links } = useFormatAdvancedNetworkData({
+        // @TODO replace with real formatData when available
```
jonathanarnault, line 19: **issue:** todo here

packages/frontend-common/formats/chart/network/NetworkView.tsx
```diff
@@ -34,13 +36,103 @@ const styles = {
...
+        );
+    }
+
+    return (
+        selectedNode?.id === (link.source! as NodeObject).id ||
+        hoveredNode?.id === (link.source! as NodeObject).id
+    );
+};
+
+export const NetworkBase = ({
```
jonathanarnault: **nit-pick:** Since this component is used is two different files, maybe it could be interesting to mov eit to its own file

packages/frontend-common/formats/chart/advancedNetwork/useFormatAdvancedNetworkData.ts
```diff
@@ -0,0 +1,113 @@
...
+                x: viz$position.x
+                    ? parseFloat(viz$position.x) * 1000
+                    : undefined,
+                y: viz$position.y
+                    ? parseFloat(viz$position.y) * 1000
+                    : undefined,
+                z: viz$position.z
+                    ? parseFloat(viz$position.z) * 1000
+                    : undefined,
+                color: viz$color
```
jonathanarnault, line 80: **question:** Shouldn't this hopok return the color formatted as RGB directly to ease color management in NetworkBase ?

jonathanarnault (changes requested):

packages/frontend-common/formats/chart/advancedNetwork/AdvancedNetworkView.tsx
ThieryMichel: Yes because we do not have a routine to generate gexf data yet.

packages/frontend-common/formats/chart/advancedNetwork/useFormatAdvancedNetworkData.ts
ThieryMichel: We need to be able to change the opacity of the color.
...
I guess I can append the transparency to the rgb string

jonathanarnault (approved):

PR #3172, opened 2025-11-20
if the field viz$size is not set lodex crash at line 70

touv:


jonathanarnault (approved):

ThieryMichel (approved):

touv:
Pour tester cf. https://lodex-dev.inist.fr/instance/nico-termino/graph/aO4D

PR #3177, opened 2025-11-21
Feat(network): Add support for colors in network graphs

jonathanarnault:
#2958


packages/frontend-common/form-fields/ArrayInput.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { Box, Button, IconButton, Stack } from '@mui/material';
+import { useCallback, type FunctionComponent } from 'react';
+import { useTranslate } from '../i18n/I18NContext';
+
+export const ArrayInput = <
+    V,
+    P extends ArrayInputComponentProps<V> = ArrayInputComponentProps<V>,
+>({
+    Component,
+    value,
```
ThieryMichel: It should be values since it is an array

packages/frontend-common/formats/chart/network3D/Network3DView.tsx
```diff
@@ -30,13 +44,15 @@ const Network3D = ({ formatData, colorSet, field }: NetworkProps) => {
...
         maxRadius,
+        colorOverrides,
     });
 
     const fieldToFilter =
         typeof field.format?.args?.fieldToFilter === 'string'
             ? field.format.args.fieldToFilter
             : null;
 
+    console.log(colorOverrides);
```
ThieryMichel: console.log spotted

packages/frontend-common/formats/chart/network/useColorOverrides.tsx
```diff
@@ -0,0 +1,28 @@
...
+    isAdvancedColorMode: boolean | undefined,
+    colorScale: ColorScaleItemMaybe[] | undefined,
+) {
+    return useMemo<Record<string, string>>(() => {
+        if (!isAdvancedColorMode) {
+            return {};
+        }
+
+        return (
+            colorScale?.reduce<Record<string, string>>((acc, item) => {
```
ThieryMichel, line 14: Shouldn't the colorScale gets cleaned / validated on save rather than when using them ?

ThieryMichel (changes requested):

packages/frontend-common/formats/chart/network/useColorOverrides.tsx
jonathanarnault: Done in PR #3205 that is stacked on top of this one

ThieryMichel (approved):

PR #3187, opened 2025-11-24
add error boundary on Format

ThieryMichel:
see https://github.com/orgs/Inist-CNRS/projects/5/views/1?pane=issue&itemId=141045931&issue=Inist-CNRS%7Clodex%7C3182

image


packages/public-app/src/Format.spec.tsx
ThieryMichel, line 1: Tried to redo a test using testing-library, but the tentative failed because of aphrodite that make jest crash

jonathanarnault (approved):

PR #3211, opened 2025-11-25
Heatmap linked data

ThieryMichel:
see https://github.com/orgs/Inist-CNRS/projects/5/views/1?pane=issue&itemId=140272819&issue=Inist-CNRS%7Clodex%7C3154

### DIFF
--- a/packages/api/src/controller/api/publishedDataset.ts
+++ b/packages/api/src/controller/api/publishedDataset.ts
@@ -145,6 +145,58 @@ export const getPage = async (ctx: any) => {
     };
 };
 
+const getFilter = ({
+    field,
+    fieldName,
+    value,
+}: {
+    field: string;
+    fieldName: string;
+    value: unknown;
+}) => {
+    if (!value) {
+        return { field };
+    }
+
+    if (Array.isArray(value)) {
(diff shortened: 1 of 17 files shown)


jonathanarnault (approved):

packages/api/src/controller/api/publishedDataset.ts
```diff
@@ -145,6 +145,58 @@ export const getPage = async (ctx: any) => {
...
+            {
+                [`versions.0.${fieldName}`]: {
+                    $elemMatch: {
+                        $elemMatch: { $eq: value },
+                    },
+                },
+            },
+        ],
+    };
+};
```
jonathanarnault, line 184: Why not using `[].concat(value)` and using only $in filter ?

jonathanarnault (changes requested):

packages/api/src/controller/api/publishedDataset.ts
ThieryMichel: It would break previous behavior, but you're right it's uselesssly complicated

jonathanarnault (approved):

PR #3253, opened 2025-12-01
Fix network viewport

ThieryMichel:
see https://github.com/orgs/Inist-CNRS/projects/5/views/1?pane=issue&itemId=142014072&issue=Inist-CNRS%7Clodex%7C3236 and https://github.com/orgs/Inist-CNRS/projects/5/views/1?filterQuery=ancienne&pane=issue&itemId=133099871&issue=Inist-CNRS%7Clodex%7C2969

### DIFF
--- a/packages/api/src/controller/api/field.ts
+++ b/packages/api/src/controller/api/field.ts
@@ -5,6 +5,7 @@ import route from 'koa-route';
 import moment from 'moment';
 import streamToString from 'stream-to-string';
 import tar from 'tar-stream';
+import { v1 as uuid } from 'uuid';
 
 import { validateField } from '../../models/field';
 import indexSearchableFields from '../../services/indexSearchableFields';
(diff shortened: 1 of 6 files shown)


packages/api/src/controller/api/field.ts
```diff
@@ -460,6 +463,18 @@ export const duplicateField = async (ctx: any) => {
...
         // save field in database
         const newField = await ctx.field.create(field);
+
+        ctx.body = await ctx.field.updateOneById(newField._id, newField);
+        await indexSearchableFields(ctx);
+        if (newField.searchable) {
+            await indexSearchableFields(ctx);
+        }
+
+        await workerQueues[ctx.tenant].add(
```
ThieryMichel: I'm triggering a replublication directly from api instead of doing it in saga on a SAVE_FIELD_SUCCESS

jonathanarnault: If the queue does not exist, this could break the field udpate

jonathanarnault (changes requested):

jonathanarnault (approved):

PR #3286, opened 2025-12-05
Feat(graphic): Add more result support to search pane

jonathanarnault:
#3200


packages/public-app/src/search/useListPublishedDatasetByFieldValue.tsx
```diff
@@ -1,41 +1,84 @@
...
-                    }),
-                });
-            },
+    const { isLoading: isListSearchResultPending, mutateAsync } = useMutation<
+        PublishedDataByFieldResponse,
+        unknown,
+        number
+    >({
+        mutationKey: ['search', 'list', JSON.stringify(filters)],
+        async mutationFn(page = 0) {
```
ThieryMichel: Why not use an infiniteQuery ?
https://tanstack.com/query/v4/docs/framework/react/guides/infinite-queries
Using a mutation to fetch data seems hackish. And you do not benefit from the query cache.

ThieryMichel (changes requested):

ThieryMichel (approved):

PR #3301, opened 2025-12-08
add secondFieldToFilter params

ThieryMichel:
see https://github.com/orgs/Inist-CNRS/projects/5/views/1?pane=issue&itemId=143746850&issue=Inist-CNRS%7Clodex%7C3295


jonathanarnault (commented):
Why not name fields `fieldToFilterSource` and `fieldToFilterTarget` as it is named in other graphs ?

jonathanarnault (approved):

PR #3329, opened 2025-12-10
Network legend asterisk

ThieryMichel:
see https://github.com/orgs/Inist-CNRS/projects/5/views/1?pane=issue&itemId=143124366&issue=Inist-CNRS%7Clodex%7C3272

image
image

### DIFF
--- a/packages/frontend-common/formats/chart/network/useColorOverrides.tsx
+++ b/packages/frontend-common/formats/chart/network/useColorOverrides.tsx
@@ -1,12 +1,19 @@
 import { useMemo } from 'react';
 import { useTranslate } from '../../../i18n/I18NContext';
 
-const matchOneOf = (patterns: string[], content: string) => {
-    return patterns.some((pattern) =>
-        new RegExp(`^${pattern.replaceAll('*', '.*')}`, 'i').test(
+export const matchOneOf = (patterns: string[], content: string) => {
+    if (!patterns || patterns.length === 0) {
+        return false;
+    }
+    return patterns.some((pattern) => {
+        console.log({
+            pattern,
+            regex: `^${pattern.replaceAll('*', '.*')}`,
+        });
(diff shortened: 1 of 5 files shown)


packages/frontend-common/formats/chart/network/useColorOverrides.tsx
```diff
@@ -1,56 +1,87 @@
 import { useMemo } from 'react';
 import { useTranslate } from '../../../i18n/I18NContext';
 
+export const matchOneOf = (patterns: string[], content: string) => {
+    if (!patterns || patterns.length === 0) {
+        return false;
+    }
+    return patterns.some((pattern) => {
+        return new RegExp(`^${pattern.replaceAll('*', '.*')}`, 'i').test(
```
jonathanarnault: Maybe a dollar in the regex ? Otherwise this wildcard `a*e` could match 'abcdefg`

```suggestion
        return new RegExp(`^${pattern.replaceAll('*', '.*')}$`, 'i').test(
```

packages/frontend-common/formats/chart/network/useColorOverrides.tsx
```diff
@@ -1,56 +1,87 @@
...
                 },
-                {
-                    colorOverrides: {},
-                    captions: defaultColor
-                        ? {
-                              [translate('other')]: defaultColor,
-                          }
-                        : {},
-                },
+                (defaultColor
```
jonathanarnault, line 55: I think that this would be better if the caption contained all matched values instead of `1 - *` this would be easier for the user to know which values are selected by the wildcard

jonathanarnault (changes requested):

packages/frontend-common/formats/chart/network/useColorOverrides.tsx
ThieryMichel: I'm afraid this would generate ginormous label.

jonathanarnault (approved):

PR #3341, opened 2025-12-11
Fix(hierarchy): Add spacing and add label sort

jonathanarnault:
#3331 

Capture d’écran du 2025-12-12 08-57-28


ThieryMichel (approved):

ThieryMichel (changes requested):
Screenshot or it did not happens 😛

ThieryMichel (approved):

PR #3346, opened 2025-12-12
Feat(tree): Show leafs as square

jonathanarnault:
#3337, #3244 

Capture d’écran du 2025-12-12 11-01-55

Capture d’écran du 2025-12-12 14-17-12

### DIFF
--- a/packages/frontend-common/formats/chart/network/NetworkBase.tsx
+++ b/packages/frontend-common/formats/chart/network/NetworkBase.tsx
@@ -143,6 +143,7 @@ type NetworkBaseProps = {
     zoomAdjustNodeSize?: boolean;
     captionTitle?: string;
     captions?: Record<string, string>;
+    displayDifferentShape?: boolean;
 };
 
 export const NetworkBase = ({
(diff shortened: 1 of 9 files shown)


packages/frontend-common/formats/chart/network/NetworkBase.tsx
```diff
@@ -425,17 +425,28 @@ export const NetworkBase = ({
...
                 ctx.beginPath();
-                ctx.arc(
-                    node.x!,
-                    node.y!,
-                    circleRadius + (zoomAdjustNodeSize ? 1 / globalScale : 1),
-                    0,
-                    2 * Math.PI,
-                    false,
-                );
+                if (node.isLeaf) {
```
ThieryMichel: SHouldn't it be done only if two fieldToFilters are configured ?

ThieryMichel (changes requested):
What about 3d network ?

packages/frontend-common/formats/chart/network/NetworkBase.tsx
jonathanarnault: No, because should be different even if fields to filter are not configured

ThieryMichel (approved):

PR #3614, opened 2026-06-10
use a local file rather than a network file

touv:


packages/workers/src/routines/distinct-ISO3166-1-alpha2-from.ini
```diff
@@ -14,20 +14,30 @@ connectionStringURI = get('connectionStringURI')
...
+location = ./src/resources/
+
+[combine/JSONParse]
+separator = *
+
+[combine/replace]
+path = id
+value = get('English name')
+path = value
+value =  get('alpha-3 code')
```
touv: alpha-2

